PR #79, opened 2015-04-21
Play 24 migration

schleichardt:


packaging.sbt
```diff
@@ -1,9 +0,0 @@
-import NativePackagerKeys._
```
schleichardt: I removed this file since we deploy to heroku and don't need this stuff right now. Also the API changed.

schleichardt:
I also included the metrics thing here in the pull request. I had to do something while waiting for travis...

schleichardt:
It looks like this:

`[debug] s.m.simple - GET / used 1 requests (1 queries, 0 commands, 7232 bytes fetched, in (56 ms))`

schleichardt:
@lauraluiz please review

yanns:
great example, you should talk about it at the play meetup ;)

lauraluiz:
@yanns +1 ;)

app/plugins/ProductionModule.java
```diff
@@ -1,75 +1,25 @@
...
-
-    public static PlayJavaSphereClient createClient() {
-        Logger.info("execute PlayJavaSphereClientProvider.get()");
-        final Config classpathConf = ConfigFactory.load();
-        final String project = classpathConf.getString("sphere.project");
-        final String clientId = classpathConf.getString("sphere.clientId");
-        final String clientSecret = classpathConf.getString("sphere.clientSecret");
-        final SphereClient sphereClient = SphereClientFactory.of().createClient(project, clientId, clientSecret);
-        return PlayJavaSphereClient.of(sphereClient);
+    public Seq<Binding<?>> bindings(Environment environment, play.api.Configuration configuration) {
```
lauraluiz: Remove namespace from `Configuration` (unless you did it for a reason).

schleichardt: Yes, this is the Scala Configuration class, not the one of the Java API.

schleichardt:
Currently two tests fail, one is only executed in travis - so I didn't see that coming locally and the other is a timeout.

app/plugins/ProductionModule.java
```diff
@@ -1,75 +1,25 @@
...
+        return seq(
```
lauraluiz: It's ugly that they require a Scala `Seq`, I hope they change it...

app/plugins/ProductionModule.java
lauraluiz: Since you are using the Scala API everywhere in this class, I didn't see it necessary. But I guess you want to really make the developer clear it is not the Java API, in which case is good ;)

schleichardt:
Okay, one failing test is related to a Play Framework bug:

https://github.com/playframework/playframework/issues/3004

schleichardt:
see also https://github.com/playframework/playframework/issues/1088

lauraluiz:
I feel it took me forever, but finally I finished reviewing! :)
Let's see what Travis says, and we can merge.

PR #80, opened 2015-04-24
Module structure

lauraluiz:
@lgiavedoni these are the changes @schleichardt and me made for the new structure. Still webapp is not the main module, so ignore it, we decided to take small steps first. Anyway, later is just a matter of moving the whole content to the webapp module.

Can you review it, please?


lauraluiz:
@schleichardt for some reason the build fails because of the routing system, but only on Travis. Any idea why?

product-catalog/conf/Products.routes
```diff
@@ -0,0 +1,2 @@
+
+GET     /       products.controllers.ProductController.index()
```
schleichardt: rename to  `product-catalog/conf/products.routes` (lowercase product)
and the error message
```
[error] /home/michael/dev/sphere-sunrise/conf/routes:7: object Routes is not a member of package products
[error] ->          /products                 products.Routes
[error] /home/michael/dev/sphere-sunrise/conf/routes:7: object Routes is not a member of package products
[error] ->          /products                 products.Routes
[error] /home/michael/dev/sphere-sunrise/conf/routes:7: not found: value handler
[error] ->          /products                 products.Routes
[error] three errors found
```

will go away

lauraluiz: Thanks! I didn't realize I had the filename uppercase.

PR #99, opened 2015-06-19
Cms service

lauraluiz:
@schleichardt @butenkor please review.
Since the architecture is still under development, there are some dummy classes, please ignore them ;)
Focus on the CmsService and the HandlebarsViewService, which is the main focus of this PR.


lauraluiz:
I'm still working a bit in the Handlebars side, though. Maybe you can wait till Monday to review, better ;)

lauraluiz:
Done. It can be reviewed.

app/controllers/ApplicationController.java
```diff
@@ -23,16 +19,12 @@
...
     }
 
     public F.Promise<Result> index() {
-        return client().execute(ProductProjectionQuery.of(CURRENT)).map(this::showProducts);
-    }
-
-    private Result showProducts(PagedQueryResult<ProductProjection> productPagedQueryResult) {
-        final List<ProductProjection> products = productPagedQueryResult.getResults();
-        return ok(products.toString());
+        return F.Promise.pure(redirect(productcatalog.controllers.routes.ProductCatalogController.pop()));
```
butenkor: Shouldn't we use ```templateService``` here?

app/controllers/ApplicationController.java
```diff
@@ -23,16 +19,12 @@
 public final class ApplicationController extends SunriseController {
 
     @Inject
-    public ApplicationController(final PlayJavaSphereClient client, final CategoryTree categoryTree, final Configuration configuration) {
-        super(client, categoryTree, configuration);
+    public ApplicationController(final PlayJavaSphereClient client, final CategoryTree categoryTree,
```
schleichardt: Maybe a parameter object can make this more smooth.

app/controllers/ApplicationController.java
```diff
@@ -23,16 +19,12 @@
...
+                                 final Configuration configuration, final TemplateService templateService, final CmsService cmsService) {
+        super(client, categoryTree, configuration, templateService, cmsService);
     }
 
     public F.Promise<Result> index() {
```
schleichardt: I think `F.Promise<Result>` should be `Result` if you just use `Promise.pure`.

app/controllers/ApplicationController.java
schleichardt: It is a redirect, it does not have any template logic.

lauraluiz: For the 303 redirection we need the inverse route (which is converted into a URL) as it is used here, which is independent from the template system used. The TemplateService only knows which template to use for the required page, it knows nothing about URLs.
Notice what I'm doing here is simply redirect the index page to the POP so that Sunrise is initially showing something. This will change once we have the home page designed.

app/controllers/ApplicationController.java
lauraluiz: Yep, I was thinking of grouping some of the services to reduce the number of parameters, but I wanted to have a clear idea of how many will we eventually have before taking a decision in that direction.

app/controllers/ApplicationController.java
lauraluiz: But then I have to apply `Promise.pure` in the `HomeController` instead (that's why I put the `pure` in `ApplicationController`). It's just a temporary method that will eventually change into a `Promise.pure`, that's why I chose to put the pure in here.

app/controllers/ApplicationController.java
schleichardt: In nightfall I had just a class ControllerDeps and never worried about it. Inside the super class it was unpacking this and provides methods to access it.

lauraluiz: Interesting, I've created a ticket https://github.com/sphereio/sphere-sunrise/issues/100 to apply it in the next PR, which by the way, has the honor to be the 100th issue ;)

app/inject/ProductionModule.java
```diff
@@ -23,7 +25,9 @@
         return seq(
                 bind(CountryCode.class).qualifiedWith("default").toInstance(defaultCountry(configuration)), // checks on start
                 bind(PlayJavaSphereClient.class).toProvider(PlayJavaSphereClientProvider.class).in(Singleton.class),
-                bind(CategoryTree.class).toProvider(CategoryTreeProvider.class)
+                bind(CategoryTree.class).toProvider(CategoryTreeProvider.class),
```
schleichardt: I think Providers are cool but there is the danger that you expect a singleton where it isn't. (at-Singleton on the provider reuses only the provider). Suppose you have multiple controllers, should everyone have its own copy of a CategoryTree, TemplateService and CmsService or should they share one instance?

`bind(Hello.class).toProvider(HelloProvider.class).in(Singleton.class) //make singleton even if class have not the annotation`

common/app/common/cms/CmsService.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface CmsService {
+
+    /**
+     * Gets the page content corresponding to the given key for a certain language.
+     * @param locale for the localized text
+     * @param pageKey identifying the page
+     * @return the promise of the page content identified by the key, in the given language
+     */
+    F.Promise<CmsPage> get(final Locale locale, final String pageKey);
```
schleichardt: What happens if a cms page is not found.

common/app/common/templates/TemplateService.java
```diff
@@ -0,0 +1,28 @@
+package common.templates;
+
+import common.pages.PageData;
+import play.twirl.api.Html;
+
+/**
+ * Service that provides HTML pages, using some kind of template system.
+ */
+public interface TemplateService {
```
schleichardt: What about calling this "Views"?

common/test/common/cms/CmsPageTest.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CmsPageTest {
+
+    @Test
+    public void getsMessage() throws Exception {
+        final CmsPage cmsPage = ((k, a) -> Optional.of("bar"));
+        assertThat(cmsPage.getOrEmpty(someMessageKey())).isEqualTo("bar");
```
schleichardt: I don't think you testing any behaviour here.

common/test/common/templates/TemplateServiceTest.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    private String someHtml() {
+        return "<html>" +
+                "<head></head>" +
+                "<body>Hello world!</body" +
+                "</html>";
+    }
+
+    private String someTemplateName() {
```
schleichardt: Why do you use a method here and not a value in the method or class?

common/app/common/cms/CmsService.java
schleichardt: I would rename it getPage.

common/app/common/templates/TemplateService.java
lauraluiz: That was the initial name but I realized it was wrong. This service only knows about templates. So you give it a template name and data to inject, and returns the HTML. It doesn't actually know about views. Who knows about views is, for example, the `ProductCatalogView`, which knows the view you want to obtain, so it calls the proper template and requests the proper data and delivers the HTML.

common/app/common/cms/CmsService.java
schleichardt: Or findPage.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -0,0 +1,37 @@
...
+public class ProductCatalogController extends SunriseController {
+
+    @Inject
+    public ProductCatalogController(final PlayJavaSphereClient client, final CategoryTree categoryTree,
+                                    final Configuration configuration, final TemplateService templateService, final CmsService cmsService) {
+        super(client, categoryTree, configuration, templateService, cmsService);
+    }
+
+    public F.Promise<Result> pop() {
+        return cmsService().get(userContext().locale(), "pop").flatMap(cms -> {
```
schleichardt: I would extract a method, so that you could write:

```java
getPage("pop").flatMap(cms -> {
```

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -0,0 +1,37 @@
...
+            final ProductOverviewPageContent pageData = new ProductOverviewPageContent(cms);
+            return view().map(view -> ok(view.productOverviewPage(pageData)));
```
schleichardt: For me is not really clear what here is going on.

common/app/common/templates/TemplateService.java
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Injects the page data into the template with the given name.
+     * @param templateName name of the template
+     * @param pageData data to be injected in the template
+     * @return string of the HTML generated with the template and the given page data
+     */
+    String fill(final String templateName, final PageData pageData);
```
schleichardt: The name fill sounds like you fill the the pageData, but it actually renders stuff. How about calling this "render"?

common/app/common/templates/TemplateService.java
schleichardt: Interesting. What about Templates?

common/test/common/templates/TemplateServiceTest.java
lauraluiz: Just a way to hide that I actually don't care about the value. The template service is dummy, it always returns the same no matter what do you pass. I could pass `"foo"` or `null` anywhere, but I thought it was easy to get distracted about the value and people would wonder there is a `null` or `"foo"` there, so I hid it behind a `some...` method.
It's just a little detail I thought it was a good idea, but if you guys think otherwise I don't have a problem removing it :P

app/controllers/ApplicationController.java
schleichardt: Ok.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
lauraluiz: Yep, I have to work on the controller side to make it easier to use, but right now I decided the controllers would be more a playground in this PR to prove the concept, in order to avoid touching too many files and scopes. Also, it will be different once we have some real controller action that reflects some use case (right now they are quite dummy). In that moment I'll see what improvements I can do.

common/app/common/templates/TemplateService.java
lauraluiz: Thanks! I was looking for a name like that and I couldn't come with it. I'll change it right away :)

app/inject/TemplateServiceProvider.java
```diff
@@ -0,0 +1,27 @@
...
+import common.templates.TemplateService;
+import play.Logger;
+
+public class TemplateServiceProvider implements Provider<TemplateService> {
+    private static final String DEFAULT_TEMPLATE_PATH = "/META-INF/resources/webjars/templates";
+    private static final String OVERRIDE_TEMPLATE_PATH = "app/assets/templates";
+
+    @Override
+    public TemplateService get() {
+        Logger.info("execute ViewServiceProvider.get()");
```
butenkor: I think this belongs more to ```Logger.debug``` level.

app/inject/ProductionModule.java
```diff
@@ -23,7 +25,9 @@
         return seq(
```
butenkor: Just wonder why we have to use Scala here?

common/app/common/cms/PlayCmsService.java
```diff
@@ -0,0 +1,38 @@
...
+import play.i18n.MessagesApi;
+import play.libs.F;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.Locale;
+
+/**
+ * Service that provides content data from Messages files as handled by Play.
+ * In order to share a common interface with other CMS, a normal message key has been split into Page Key and Message Key.
```
butenkor: Where do you split?

common/app/common/cms/CmsService.java
butenkor: This maybe not be enough. For example sometimes you need to pass some flags like: ```isPreview``` or any other helper parameter.

butenkor: Should not that be an ```F.Promise<Option<CmsPage>>```?

common/app/common/controllers/SunriseController.java
```diff
@@ -18,17 +20,30 @@
...
         this.countryOperations = CountryOperations.of(configuration);
+        this.templateService = templateService;
+        this.cmsService = cmsService;
     }
 
     protected final CategoryTree categories() {
         return categoryTree;
     }
 
+    public TemplateService viewService() {
```
butenkor: To omit confusion maybe we should stick to single name ```templateService``` or ```viewService```?

common/app/common/cms/PlayCmsService.java
lauraluiz: We have `PlayCmsService.getPage(locale, pageKey)`, which returns a CmsPage, and then `PlayCmsPage.get(messageKey)`. I'll explain it better in the Javadoc.

app/inject/ProductionModule.java
lauraluiz: Unfortunately, it's what Play forces to use. You can check the documentation about it: https://www.playframework.com/documentation/2.4.x/JavaDependencyInjection (section Play Libraries).

app/inject/TemplateServiceProvider.java
lauraluiz: It was inherited code I copied it from. I'll change it in all other providers as well.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -0,0 +1,37 @@
...
+import play.libs.F;
+import play.mvc.Result;
+import productcatalog.models.ProductOverviewPageContent;
+import productcatalog.templates.ProductCatalogView;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class ProductCatalogController extends SunriseController {
```
butenkor: Why not well known ProductOverviewController (also called POP)? :)

lauraluiz: This name is so far simply a dummy placeholder for experimenting with the CMS/Pages/Templates (so, views side). Initially it was only *the* controller of the ProductCatalog module, that's where the name comes from. I wouldn't worry that much about the controller side, since I'm using it as a playground. In the next task I will start seriously with the POP and therefore with the full stack controllers/models/views :P

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
butenkor: For performance reasons we never fetch from CMS service twice. So first call should contains all the required CMS data.

common/app/common/cms/CmsService.java
lauraluiz: Done, changed to getPage and thrown an exception.

common/app/common/pages/PageHeader.java
```diff
@@ -0,0 +1,31 @@
+package common.pages;
+
+import common.cms.CmsPage;
+
+public class PageHeader {
```
butenkor: It may be more flexible to have just CmsPage and some container with all CMS elements and helpers to extract data by name/code.

app/inject/TemplateServiceProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.github.jknack.handlebars.io.ClassPathTemplateLoader;
+import com.github.jknack.handlebars.io.FileTemplateLoader;
+import com.google.inject.Provider;
+import common.templates.HandlebarsTemplateService;
+import common.templates.TemplateService;
+import play.Logger;
+
+public class TemplateServiceProvider implements Provider<TemplateService> {
+    private static final String DEFAULT_TEMPLATE_PATH = "/META-INF/resources/webjars/templates";
```
butenkor: Do we need default path because of webjars plugin?

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
lauraluiz: What about https://github.com/sphereio/sphere-sunrise/commit/afcb8f5a3dd17ae6b9b3550445eae8972e639f44 ?

common/app/common/cms/CmsService.java
lauraluiz: Yes, probably we'll need more optional arguments, but I would need to implement the Contentful implementation first to see a better overview. But I think as a first draft for the POC is good enough, the additional changes won't be that costly.

About the `Option`, for the page I think it's more adequate to throw an exception if the page is not found. There's not many things you can do without the content. A different approach is taken for a single message that cannot be found, there an `Optional` is returned.

common/app/common/controllers/SunriseController.java
lauraluiz: Yes, this was changed already this morning: https://github.com/sphereio/sphere-sunrise/commit/c7b728ea8795c3407c94e15e9876206b752d1606
The refactoring didn't take into account this method.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
lauraluiz: That will be something to take into account when the controller is implemented, but that doesn't affect the CMS implementation.

common/app/common/pages/PageHeader.java
lauraluiz: PageHeader has nothing to do with CMS. PageHeader (as well as PageContent, PageFooter, etc.) is about the data that has to be injected in a template. This data can come from CMS, from SPHERE or from any other external service. Therefore it is not a wrapper, but rather a user of the CMS.

app/inject/TemplateServiceProvider.java
lauraluiz: Not exactly. Handlebars needs a path where to find the templates, which in this case it is looking for them inside the sphere-sunrise-design webjar in the classpath. The webjar plugin is only useful to find the public assets of the website into the webjar.

app/inject/TemplateServiceProvider.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final String OVERRIDE_TEMPLATE_PATH = "app/assets/templates";
+
+    @Override
+    public TemplateService get() {
+        Logger.debug("execute ViewServiceProvider.get()");
```
butenkor: Should not be ```TemplateServiceProvider.get()```?

lauraluiz: Good catch :)

common/app/common/pages/PageHeader.java
butenkor: Then its more like a helper for producing well structured JSON for handlebars templates?

lauraluiz: Yes, but not only for handlebars, but for any template. We were already using this same structure for Scala templates, except that now the complexity has increased due to the CMS and Template system.

app/inject/TemplateServiceProvider.java
schleichardt: Why not `getClass().getSimpleName() + ".get()"`?

lauraluiz:
So @schleichardt @butenkor can I merge? Any stuff I missed?

butenkor:
From my side OK.

schleichardt:
ok

PR #106, opened 2015-07-10
Template rendering

peter-gerhard:


common/app/common/prices/PriceFinder.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private boolean priceHasChannel(Price price, Reference<Channel> channel) {
+        return price.getChannel().map(c -> c.hasSameIdAs(channel)).orElse(false);
+    }
+
+    private boolean priceHasValidDate(Price price, ZonedDateTime date) {
+        final boolean toEarly = price.getValidFrom().map(from -> from.isAfter(date)).orElse(false);
+        final boolean toLate = price.getValidUntil().map(until -> until.isBefore(date)).orElse(false);
+
+        return !toEarly & !toLate;
```
schleichardt: Why &?

common/test/common/prices/PriceFinderTest.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Optional;
+
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PriceFinderTest {
+
+    private final PriceFinder priceFinder = PriceFinder.of();
+
+    private final CurrencyContext eurContext = CurrencyContextBuilder.of("EUR").build();
```
schleichardt: Can you please use a more idiomatic way to initialize currencies?

`Monetary.getCurrency("EUR")`

I think currency context is not needed. Could you replace it, in the whole project?

http://sphereio.github.io/sphere-jvm-sdk/javadoc/master/io/sphere/sdk/meta/FormattingDocumentation.html#creating-money-values

common/app/common/prices/PriceFinder.java
```diff
@@ -0,0 +1,112 @@
...
+     country, customer group and channel.
+     customer group and channel
+     customer group and country
+     customer group
+     country and channel
+     channel
+     country
+     any left
+     */
+    public Optional<Price> findPrice(final Collection<Price> prices,
```
schleichardt: I would suggest to initialize a pricefinder with final CurrencyUnit currency,
+                                     final CountryCode country,
+                                     final java.util.Optional<Reference<CustomerGroup>> customerGroup,
+                                     final java.util.Optional<Reference<Channel>> channel,
+                                     final ZonedDateTime userTime

So you can easily reuse it for multiple products, for example on the pop.

So the price finding has the signature: 
`public Optional<Price> findPrice(final List<Price> prices);`
I am not sure if we can use Collection, since a Set would may produce weird results.

common/app/common/prices/PriceScope.java
```diff
@@ -0,0 +1,27 @@
...
+import com.neovisionaries.i18n.CountryCode;
+import io.sphere.sdk.channels.Channel;
+import io.sphere.sdk.customergroups.CustomerGroup;
+import io.sphere.sdk.models.Reference;
+
+import javax.money.CurrencyUnit;
+import java.time.ZonedDateTime;
+import java.util.Optional;
+
+public class PriceScope {
```
schleichardt: Is it required for the class to have public access? I think it should be in package scope since it is a helper class for PriceFinder.

common/app/common/prices/PriceScopeBuilder.java
```diff
@@ -0,0 +1,66 @@
+package common.prices;
```
schleichardt: Maybe package scope like PriceScope.

PR #112, opened 2015-07-22
Use application.conf configuration for the template handlebars paths.…

schleichardt:
@peter-gerhard please review


peter-gerhard:
:+1:

Issue #144, opened 2015-08-03
Have design JSON as fallback context


lauraluiz:
To avoid using localization when it is actually not required.

PR #146, opened 2015-08-06
Fallback context

lauraluiz:
@schleichardt @peter-gerhard @butenkor review please

With this functionality the data coming from the Page classes is combined with other sources, doing a fallback effect. In particular, we are using this to use the JSON coming from the sphere-sunrise-design as a fallback when we still don't have the feature implemented in Sunrise, in order to present a filled template all the time.

Some issues:
- Currently those parts of the template coming using a {{#with }} are not being rendered for some reason. Will have to report to the dev of Handlebars Java.
- I chose to transform the JSON file to a Map because Handlebars Java only allows Maps for the `combine` function. I've tried to pass the JSON as parent context and works the same, but I feel combine is better oriented to the functionality we are using, while parent context is mostly used to save the context on nested data. Probably inside uses the same functionality, but it is a semantic issue.
- I'm using the TemplateLoader as a way to configure the path, so that we use the same system as we do with the Handlebars TemplateLoaders. Actually the JSON file must be called the same and be equivalent, except for the extension, so technically is not that wrong. If you think it's better to change it, please suggest a better option and I'll change it.

### DIFF
--- a/common/app/common/templates/HandlebarsTemplateService.java
+++ b/common/app/common/templates/HandlebarsTemplateService.java
@@ -1,26 +1,38 @@
 package common.templates;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.jknack.handlebars.Context;
 import com.github.jknack.handlebars.Handlebars;
 import com.github.jknack.handlebars.Template;
+import com.github.jknack.handlebars.context.MapValueResolver;
+import com.github.jknack.handlebars.io.ClassPathTemplateLoader;
 import com.github.jknack.handlebars.io.TemplateLoader;
+
 import common.pages.PageData;
 import play.Logger;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Collections.emptyList;
 
 public final class HandlebarsTemplateService implements TemplateService {
     private final Handlebars handlebars;
+    private final List<TemplateLoader> fallbackContexts;
 
-    private HandlebarsTemplateService(final Handlebars handlebars) {
+    private HandlebarsTemplateService(final Handlebars handlebars, final List<TemplateLoader> fallbackContexts) {
         this.handlebars = handlebars;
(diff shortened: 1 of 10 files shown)


common/app/common/templates/HandlebarsTemplateService.java
```diff
@@ -42,17 +59,38 @@ private Template compileTemplate(final String templateName) {
...
     }
 
-    private Context buildContext(final PageData pageData) {
+    private Context buildContext(final PageData pageData, final String templateName) {
         // TODO Use resolver with cache on production
-        return Context.newBuilder(pageData)
-                .resolver(NonCachedJavaBeanValueResolver.INSTANCE)
-                .build();
+        Context.Builder builder = Context.newBuilder(pageData)
+                .resolver(NonCachedJavaBeanValueResolver.INSTANCE, MapValueResolver.INSTANCE);
```
schleichardt: Do you use in prod the cached resolver?

schleichardt:
:+1:

common/app/common/templates/HandlebarsTemplateService.java
lauraluiz: No, you can see the TODO right above this commit :P
Actually, I'll create the task so I don't forget :)

lauraluiz: https://github.com/sphereio/sphere-sunrise/issues/147

lauraluiz:
solves https://github.com/sphereio/sphere-sunrise/issues/144

lauraluiz:
I will proceed to merge it, as it is a bit old this PR already.
Any objection?

peter-gerhard:
sry lgtm :+1:

schleichardt:
:+1:

lauraluiz:
@schleichardt I already had your thumbsups, but thanks for emphasize it :D ;)

schleichardt:
:+1:

butenkor:
Bit late but looks good :)

schleichardt:
:+1:

lauraluiz:
:+1: ;)

PR #151, opened 2015-08-12
Pdp

peter-gerhard:


common/app/common/controllers/SunriseController.java
```diff
@@ -32,7 +34,7 @@
     protected SunriseController(final ControllerDependency controllerDependency) {
         super(controllerDependency.sphere());
         // TODO Fill it properly
-        final UserContext userContext = UserContext.of(GERMAN, emptyList(), DE, ZoneId.of("Europe/Berlin"));
+        final UserContext userContext = UserContext.of(DE, GERMAN, emptyList(), ZoneId.of("Europe/Berlin"), Reference.of(Zone.typeId(), "f77ddfd4-af5b-471a-89c5-9a40d8a7ab88"));
```
butenkor: Why it is hardcoded and not taken from application.conf?

common/app/common/pages/BagItemData.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private final int numItems;
+
+    public BagItemData(final int numItems) {
+        this.numItems = numItems;
+    }
+
+    public List<SelectableData> getBagItems() {
+        final Stream<SelectableData> defaultItemStream = Stream.of(bagItemDefaultItem());
+        final Stream<SelectableData> bla = IntStream.range(2, numItems).mapToObj(this::bagItemToSelectableItem);
```
butenkor: why bla? :)

common/app/common/pages/ImageData.java
```diff
@@ -0,0 +1,36 @@
...
+    public String getBigImage() {
+        return nullIfEmpty(image.getUrl());
+    }
+
+    public String getPlaceholderImage() {
+        return "http://placehold.it/350x150";
+    }
+
+    private String nullIfEmpty(final String s) {
+        if(s.isEmpty())
```
butenkor: can ```s``` be null?

app/inject/TemplateServiceProvider.java
```diff
@@ -35,7 +35,7 @@ public TemplateService get() {
                     } else if ("file".equals(type)) {
                         return new FileTemplateLoader(path);
                     } else {
-                        throw new RuntimeException("cannot build template loader for " + loaderConfig);
+                        throw new RuntimeException("cannot create template loader for " + loaderConfig);
```
schleichardt: When do you use SunriseInitializationException and when RuntimeException?

common/app/common/controllers/SunriseController.java
schleichardt: The user context should not depend on application.conf.

schleichardt: A hardcoded ID will make it fail in another project - like on my PC or Sunrise Production?

common/app/common/pages/BagItemData.java
```diff
@@ -0,0 +1,31 @@
+package common.pages;
+
+import java.util.List;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+
+public class BagItemData {
```
schleichardt: I would appreciate a little Javadoc comment what class is about.

Is the bag the same as the cart?

common/app/common/pages/ImageData.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public String getBigImage() {
+        return nullIfEmpty(image.getUrl());
+    }
+
+    public String getPlaceholderImage() {
+        return "http://placehold.it/350x150";
+    }
+
+    private String nullIfEmpty(final String s) {
```
schleichardt: I think StringUtils is just great for this and included in the classpath:

http://commons.apache.org/proper/commons-lang/javadocs/api-3.4/org/apache/commons/lang3/StringUtils.html#stripToNull(java.lang.String)

schleichardt:
Optional: Every class you introduced or changed that inherits implicitly from Object, could inherit Base, so we have a cool default for `toString` and `equals` and `hashCode`.

common/app/common/pages/RatingData.java
```diff
@@ -0,0 +1,24 @@
...
+public class RatingData {
+    private final CmsPage cms;
+
+    public RatingData(final CmsPage cms) {
+        this.cms = cms;
+    }
+
+    public List<SelectableData> getRating() {
+        return asList(
+                new SelectableData("5 Stars", "5", cms.getOrEmpty("ratingFiveStarText"), "", false),
```
schleichardt: @lauraluiz I think those translations should *not* come from the CMS but from the messages.

common/app/common/utils/Translator.java
```diff
@@ -0,0 +1,29 @@
...
+        this.userLanguage = userLanguage;
+        this.userFallbackLanguages = userFallbackLanguages;
+        this.projectLanguages = projectLanguages;
+    }
+
+    public static Translator of(final Locale userLanguage, final List<Locale> userFallbackLanguages, final List<Locale> projectLanguages) {
+        return new Translator(userLanguage, userFallbackLanguages, projectLanguages);
+    }
+
+    public String translate(final LocalizedStrings localizedStrings) {
```
schleichardt: It would be nice to have Javadoc here. It looks like you return an empty String on not found, right?

schleichardt: I think the method name does not match with the task of it, the localizedString already contains translated strings, what you do is find the best possible translation.

schleichardt: I would expect a class TranslatorTest which shows me the spec.

common/test/common/JsonUtils.java
```diff
@@ -0,0 +1,19 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+
+public class JsonUtils {
+
+    public static JsonNode readJsonNodeFromResource(final String resourcePath) throws IOException {
+        final ObjectMapper mapper = new ObjectMapper();
+        try(final InputStream resourceAsStream = Thread.currentThread().getContextClassLoader().getResourceAsStream(resourcePath)) {
```
schleichardt: cool :+1:  that you use this Java 7 sugar.

common/test/common/JsonUtils.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+
+public class JsonUtils {
+
+    public static JsonNode readJsonNodeFromResource(final String resourcePath) throws IOException {
+        final ObjectMapper mapper = new ObjectMapper();
```
schleichardt: I would not create per method call the Object mapper. Use a static member and nether give that mapper to the outside world.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -1,41 +1,109 @@
...
+        });
+    }
+
+    private F.Promise<Result> pdpx(final ProductProjection product, final ProductVariant variant) {
+        final F.Promise<List<ProductProjection>> suggestionPromise = productService.getSuggestions(categoryService.getSiblingCategories(product), NUM_SUGGESTIONS);
+        final List<RichShippingRate> shippingRates = shippingMethodService.getShippingRates(context().user().zone());
+        final List<Category> breadcrumbs = categoryService.getBreadCrumbCategories(product);
+
+        return suggestionPromise.flatMap(suggestions -> withCms("pdp", cms -> {
+            final Translator translator = Translator.of(context().user().language(), context().user().fallbackLanguages(), context().project().languages());
```
schleichardt: I would extract a method `getTranslator()` since you only use context to initialize it.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -1,41 +1,109 @@
...
+            final PriceFinder priceFinder = PriceFinder.of(context().user());
+            final PriceFormatter priceFormatter = PriceFormatter.of(GERMAN);
```
schleichardt: This could be initialized with the context, too.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -1,41 +1,109 @@
...
+        final F.Promise<Optional<ProductProjection>> productOptPromise = productService.searchProductBySlug(GERMAN, slug);
+
+        return productOptPromise.flatMap(productOptional -> {
+            final Optional<F.Promise<Result>> resultPromise = productOptional.flatMap(product ->
+                    productService.findVariantBySku(product, sku).map(variant -> pdpx(product, variant)));
+            return resultPromise.orElse(F.Promise.pure(notFound()));
+        });
+    }
+
+    private F.Promise<Result> pdpx(final ProductProjection product, final ProductVariant variant) {
```
schleichardt: I would break the method into two, since you have two abstraction layers. One fetches data and the other one formates data.

In addition I would not use `withCms`. `cms` is just another data source and the automatic resource management usage still is not necessary.

product-catalog/app/productcatalog/exceptions/ProductNotFoundException.java
```diff
@@ -0,0 +1,3 @@
+package productcatalog.exceptions;
+
+public class ProductNotFoundException extends Exception {}
```
schleichardt: For a really good exception it would be cool to propagate some context. Maybe you want to get the product for a certain sku or a certain slug.

schleichardt: This could just be part of the message so no extra fields are required.

product-catalog/app/productcatalog/models/RichShippingRate.java
```diff
@@ -0,0 +1,15 @@
+package productcatalog.models;
+
+import io.sphere.sdk.models.Base;
+import io.sphere.sdk.shippingmethods.ShippingRate;
+
+public class RichShippingRate extends Base {
```
schleichardt: :moneybag::moneybag::moneybag:ShippingRate

@lauraluiz Do we have some naming conventions? ShippingMethodData?

product-catalog/app/productcatalog/services/ProductProjectionService.java
```diff
@@ -0,0 +1,74 @@
...
+import io.sphere.sdk.products.queries.ProductProjectionQuery;
+import io.sphere.sdk.products.search.ProductProjectionSearch;
+import io.sphere.sdk.queries.PagedQueryResult;
+import io.sphere.sdk.queries.PagedResult;
+import play.libs.F;
+
+import javax.inject.Inject;
+import java.util.*;
+
+public class ProductProjectionService {
```
schleichardt: Normally we split every Service in an interface and an Impl class. The interface contains verbose Javadoc what it does.

product-catalog/app/productcatalog/services/ProductProjectionService.java
```diff
@@ -0,0 +1,74 @@
...
+                .map(PagedQueryResult::head);
+    }
+
+    public Optional<ProductVariant> findVariantBySku(final ProductProjection product, final String sku) {
+        return product.getAllVariants().stream()
+                .filter(variant -> variantHasSku(variant, sku))
+                .findFirst();
+    }
+
+    public F.Promise<List<ProductProjection>> getSuggestions(final List<Category> categories, final int numSuggestions) {
```
schleichardt: For testing I would split this method into two.
One method has this signature and the other one has the same signatures plus a parameter of type Random and is called by the other one, so your can test the logic in a pure functional way since you can in tests pass a certain Random as parameter.

product-catalog/app/productcatalog/services/ProductProjectionService.java
```diff
@@ -0,0 +1,74 @@
...
+    private boolean variantHasSku(final ProductVariant variant, final String sku) {
+        return variant.getSku().map(variantSku -> variantSku.equals(sku)).orElse(false);
+    }
+
+    private <T> List<T> pickNRandom(final List<T> elements, final int n) {
+        if(elements.size() < n)
+            return pickNRandom(elements, elements.size());
+
+        final List<T> picked = new ArrayList<>();
+        final Random random = new Random();
```
schleichardt: I would use a Random field in the class level. Since requests at same time may be not that random...

product-catalog/app/productcatalog/services/ProductProjectionService.java
```diff
@@ -0,0 +1,74 @@
...
+
+        for(int i = 0; i < n; i++)
```
schleichardt: I would always add curly braces around loops and conditional statements. This is less error prone.

product-catalog/app/productcatalog/services/ProductProjectionService.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private boolean variantHasSku(final ProductVariant variant, final String sku) {
+        return variant.getSku().map(variantSku -> variantSku.equals(sku)).orElse(false);
+    }
+
+    private <T> List<T> pickNRandom(final List<T> elements, final int n) {
+        if(elements.size() < n)
+            return pickNRandom(elements, elements.size());
+
+        final List<T> picked = new ArrayList<>();
```
schleichardt: If n is the size of elements to pick then you can configure the array list to have the initial capacity n:

```new ArrayList<>(n);```

File: product-catalog/app/productcatalog/services/ProductProjectionService.java [hunk 21]
```diff
@@ -0,0 +1,74 @@
...
+        final Random random = new Random();
+
+        for(int i = 0; i < n; i++)
+            pick(elements, picked, random.nextInt(elements.size()));
+
+        return picked;
+    }
+
+    private <T> void pick(final List<T> elements, final List<T> picked, int index) {
```
schleichardt: I would inline (which is kind of remove) this method since it has output parameters/does side effects.

schleichardt: Mutable state is perfectly fine to use inside a method, but outside it is for me source of errors and confusion.

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private final Translator translator = Translator.of(german, emptyList(), emptyList());
+    private final PriceFinder priceFinder = PriceFinder.of(eur, de, Optional.empty(), Optional.empty(), todayDe);
+    private final PriceFormatter priceFormatter = PriceFormatter.of(german);
+
+    private final CategoryTree categories = CategoryTree.of(CategoryUtils.getQueryResult("categoryQueryResult.json").getResults());
+    private final List<ProductProjection> products = getQueryResult("productProjectionQueryResult.json").getResults();
+
+    @Test
+    public void testStaticJson() throws IOException {
```
schleichardt: Starting a test method with the name "test" is a JUnit 3 thing, in JUnit 4 this should be avoided. IntelliJ IDEA creates always this prefix but this is not according JUnit 4 conventions. 

You can change that in your IDE:
Settings | Editor | File and Code Templates | Code tab and then "JUnit 4 Test Method".

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
```diff
@@ -0,0 +1,134 @@
...
+        final CmsPage cms = (messageKey, args) -> Optional.of(messageKey);
+        final PdpStaticData pdpStaticData = new PdpStaticData(cms);
+
+        final JsonNode expected = readJsonNodeFromResource("pdpStaticData.json");
+        final JsonNode result = toJsonNode(pdpStaticData);
+
+        assertThat(result).isEqualTo(expected);
```
schleichardt: I think the name of the test is not descriptive enough but since it is hard to write good stuff in camel case I would add here stuff *why* you check for equality.

```
assertThat(result)
.overridingErrorMessage("The Sunrise classes should produce the same JSON output as delivered by Sunrise Design.")
.isEqualTo(expected);
```

@lauraluiz maybe you can improve the sentence...

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Test
+    public void testBreadcrumbJson() throws IOException {
```
schleichardt: Cool, one can see he arrange, act, assert pattern.

common/app/common/controllers/SunriseController.java
butenkor: @schleichardt why it should not depend on application.conf? In my opinion its better to have configurables and some fallback defaults.

common/app/common/pages/BagItemData.java
butenkor: :+1: Generally i would love to seem more java doc.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -1,41 +1,109 @@
...
+        final ProductThumbnailDataFactory thumbnailDataFactory = ProductThumbnailDataFactory.of(translator, priceFinder, priceFormatter);
+
+        final String additionalTitle = "";
+        final List<ProductThumbnailData> productList = products.stream().map(thumbnailDataFactory::create).collect(toList());
+
+        return new ProductOverviewPageContent(additionalTitle, productList);
+    }
+
+    public F.Promise<Result> pdp(final String slug, final String sku) {
+        final F.Promise<Optional<ProductProjection>> productOptPromise = productService.searchProductBySlug(GERMAN, slug);
```
butenkor: Why GERMAN is hardcoded? You could use current locale. Demo will not support english yet?

common/app/common/controllers/SunriseController.java
schleichardt: The user context is the context from the user currently visiting the shop, as a result this is a request scoped information and not shop wide information.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -1,41 +1,109 @@
...
+
+        return productOptPromise.flatMap(productOptional -> {
+            final Optional<F.Promise<Result>> resultPromise = productOptional.flatMap(product ->
+                    productService.findVariantBySku(product, sku).map(variant -> pdpx(product, variant)));
```
butenkor: Why you do 2 requests (search by slug and then search by sku)? Why not from beginning by sku?

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
butenkor: what does pdpx stand for?

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
```diff
@@ -1,41 +1,109 @@
...
+            return resultPromise.orElse(F.Promise.pure(notFound()));
+        });
+    }
+
+    private F.Promise<Result> pdpx(final ProductProjection product, final ProductVariant variant) {
+        final F.Promise<List<ProductProjection>> suggestionPromise = productService.getSuggestions(categoryService.getSiblingCategories(product), NUM_SUGGESTIONS);
```
butenkor: Could we rename ```getSuggestions``` to something else like ```getSimilarProducts, getBySiblingCategories, getRandomProductsBy```? We will need suggestions naming later for the search and terms suggestion: http://dev.sphere.io/http-api-projects-products-suggestions.html

product-catalog/app/productcatalog/models/RichShippingRate.java
butenkor: :+1: for consitent naming convention. If we wrap jvm's ```ShippingRate``` then i would vote for ```CustomShippingRate ``` or maybe ```ShippingRateData```

product-catalog/app/productcatalog/pages/ProductDetailPageContent.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public ProductData getProduct() {
+        return productData;
+    }
+
+    public List<ProductThumbnailData> getSuggestions() {
+        return suggestions;
+    }
+
+    public List<ShippingRateData> getDelivery() {
```
butenkor: its a list so maybe ```getShippingRates()``` is more appropriate name?

product-catalog/app/productcatalog/pages/ProductDetailPageContent.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public PdpStaticData getStatic() {
+        return staticData;
+    }
+
+    public String additionalTitle() {
+        return additionalTitle;
+    }
+
+    public List<LinkData> getBreadcrumb() {
```
butenkor: Maybe better:
```
getBreadcrumbLinks()
```

product-catalog/app/productcatalog/services/ProductProjectionService.java
butenkor: :+1:

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
butenkor: :+1:

common/app/common/controllers/SunriseController.java
butenkor: Agree but currently its neither configuration nor current user data from request scope.

app/inject/ShippingMethodProvider.java
```diff
@@ -0,0 +1,35 @@
...
+public class ShippingMethodProvider implements Provider<ShippingMethods> {
+    private final SphereClient client;
+
+    @Inject
+    private ShippingMethodProvider(final SphereClient client) {
+        this.client = client;
+    }
+
+    @Override
+    public ShippingMethods get() {
```
lauraluiz: @peter-gerhard I forgot why you needed to keep the shipping methods locally, but I just remembered we have a nice functionality on the Shipping Methods endpoint to find the appropriate shipping for a cart: http://dev.sphere.io/http-api-projects-shippingMethods.html#shipping-methods-for-cart
I think using that convenient function in Sunrise would be a nice thing to have. So please, refresh my memory :P

app/inject/TemplateServiceProvider.java
lauraluiz: I created SunriseInitializationException for problems on fetching configuration and data needed to start the application, so I think this can be a candidate to use it.

common/app/common/contexts/UserContext.java
```diff
@@ -16,60 +17,61 @@
...
-                        final CountryCode country, final ZoneId zoneId,
-                        final Reference<CustomerGroup> customerGroup, final Reference<Channel> channel) {
-        this.language = language;
+    private UserContext(final CountryCode country, final Locale language, final List<Locale> fallbackLanguages, final ZoneId zoneId, final Reference<Zone> zone, final Reference<CustomerGroup> customerGroup, final Reference<Channel> channel) {
         this.country = country;
-        this.zoneId = zoneId;
+        this.language = language;
         this.fallbackLanguages = fallbackLanguages;
+        this.zoneId = zoneId;
+        this.zone = zone;
```
lauraluiz: Does zone (reference) contain zoneID or are they completely different things?

common/app/common/controllers/SunriseController.java
lauraluiz: It is hardcoded just because we are not going to implement the functionality that changes (most of) this information in the MVP, so so far it stays there hardcoded. So we don't forget, I've created an issue: https://github.com/sphereio/sphere-sunrise/issues/155

lauraluiz: And yes, it should come from the user. At most, we can have some default configuration.

common/app/common/pages/ImageData.java
lauraluiz: Not previous to M16, and since it is not Optional, I don't think it is nullable in M16.

common/app/common/pages/ImageData.java
```diff
@@ -0,0 +1,36 @@
...
+    public static ImageData of(final Image image) {
+        return new ImageData(image);
+    }
+
+    public String getThumbImage() {
+        return nullIfEmpty(image.getUrl());
+    }
+
+    public String getBigImage() {
+        return nullIfEmpty(image.getUrl());
```
lauraluiz: What's the benefit of returning `null` instead of empty string to the template?

app/inject/ShippingMethodProvider.java
schleichardt: I think this endpoint was not available in the SDK.

common/app/common/pages/RatingData.java
lauraluiz: @trangng I have actually a doubt here. In the final shape of Sunrise, what kind of messages should go to CMS and what kind should not be modifiable by the merchant? Last time I asked it was generally "everything", but probably there will be things we don't want to. We need to know exactly how to draw the line.

trangng: @lauraluiz is this the headline of rating? If it is then it can me in the message files, because it's a content which won't change time to time

lauraluiz: @trangng but do we have any kind of list we can use to decide what is CMS content and what is not? Besides asking you :P

common/app/common/utils/PriceFormatter.java
```diff
@@ -1,10 +1,22 @@
 package common.utils;
 
-import common.contexts.AppContext;
-
 import javax.money.MonetaryAmount;
+import javax.money.format.MonetaryFormats;
+import java.util.Locale;
+
+public class PriceFormatter {
```
lauraluiz: Remember to document with Javadoc all classes and methods that are not from Pages (those I think they are very self-explained).

product-catalog/app/productcatalog/models/RichShippingRate.java
lauraluiz: There is no written convention yet, as we need first to agree on the convention and for that I was waiting to have a complete full functionality to have the whole picture. The `FooData` naming I started using for the models that are sent to the templates, but I think this class is not one of them. For classes that are coming from CTP but we are kind of specific for this shop, I started calling them `ShopFoo`, so it would be `ShopShippingRate` (the `Custom` prefix started becoming weird with Customers, Custom Objects and so on :P). But as I said, nothing is established and it will probably change it the near future.

product-catalog/app/productcatalog/pages/ProductDetailPageContent.java
lauraluiz: `breadcrumb` is the expected name defined by the JSON data:
https://github.com/sphereio/sphere-sunrise-design/blob/master/input/templates/pdp.json#L868

The current value resolver looks for methods with a prefix `getBreadcrumb` or `isBreadcrumb`. So it needs to be called like that. We can change it in the JSON data, but there it fits better simply `breadcrumb`.

product-catalog/app/productcatalog/pages/ProductDetailPageContent.java
lauraluiz: Same as before. Delivery is the name of the section in the template. The implementation details are not exposed there, so this class acts as a interpreter between the design and the implementation.

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
lauraluiz: I think your sentence is good. But even though here I think the overrideErrorMessage is necessary, I would still change the test name for something more descriptive.

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private final CategoryTree categories = CategoryTree.of(CategoryUtils.getQueryResult("categoryQueryResult.json").getResults());
+    private final List<ProductProjection> products = getQueryResult("productProjectionQueryResult.json").getResults();
+
+    @Test
+    public void testStaticJson() throws IOException {
+        final CmsPage cms = (messageKey, args) -> Optional.of(messageKey);
+        final PdpStaticData pdpStaticData = new PdpStaticData(cms);
+
+        final JsonNode expected = readJsonNodeFromResource("pdpStaticData.json");
```
lauraluiz: My idea was to get this JSON from the webjar directly and (in this case) access the field you want to compare. If you create your own JSON files then we have to update these JSON files every time the template is changing, which destroys the purpose of having an external library so that we can forget about frontend changes.

product-catalog/test/productcatalog/pages/ProductDetailPageContentTest.java
```diff
@@ -0,0 +1,134 @@
...
+import static common.products.ProductUtils.getProductById;
+import static common.products.ProductUtils.getQueryResult;
+import static io.sphere.sdk.json.JsonUtils.readObjectFromResource;
+import static io.sphere.sdk.json.JsonUtils.toJsonNode;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ProductDetailPageContentTest {
```
lauraluiz: I feel in here you are actually testing the factories and not so much this `ProductDetailPageContent`. If that's the case, the tests should go into different test classes.

common/app/common/pages/RatingData.java
trangng: @laura yes, @CT-Szilvia is working on it and can send you the link in confluence and upload info in an issue in github

product-catalog/test/productcatalog/services/CategoryServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final CategoryTree categories = CategoryTree.of(CategoryUtils.getQueryResult("categoryQueryResult.json").getResults());
+    private final List<ProductProjection> products = ProductUtils.getQueryResult("productProjectionQueryResult.json").getResults();
+    private final CategoryService categrieService = new CategoryService(categories);
+
+    @Test
+    public void getSiblingCategories() throws ProductNotFoundException {
+        final ProductProjection suttonBag = products.stream()
+                .filter(projection -> projection.getId().equals("254f1c0e-67bc-4aa6-992d-9a0fea1846b5"))
+                .findFirst().get();
```
lauraluiz: I would extract this method that searches for the product in both tests, since they way you obtain the product is not of interest for the test and it distracts from the test goal.

product-catalog/app/productcatalog/pages/ProductDetailPageContent.java
butenkor: I think it is confusing here and also in JSON as name suggests it is single value but it is a list. You also do not name a variable ```variant``` or ```product``` if it returns an Array.

product-catalog/test/productcatalog/services/CategoryServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+        final Category clutches = categories.findById("a9c9ebd8-e6ff-41a6-be8e-baa07888c9bd").get();
+        final Category satchels = categories.findById("30d79426-a17a-4e63-867e-ec31a1a33416").get();
+        final Category shoppers = categories.findById("bd83e288-77de-4c3a-a26c-8384af715bbb").get();
+        final Category handBags = categories.findById("9a584ee8-a45a-44e8-b9ec-e11439084687").get();
+        final Category wallets = categories.findById("d2f9a2da-db3e-4ee8-8192-134ebbc7fe4a").get();
+        final Category backpacks = categories.findById("46249239-8f0f-48a9-b0a0-d29b37fc617f").get();
+        final Category shoulderBags = categories.findById("8e052705-7810-4528-ba77-00094b87a69a").get();
```
lauraluiz: Same for all the categories. Actually, I would write the test showing:
- AssertThat product has X category
- AssertThat X category has Y,Z,W siblings
- AssertThat the method returns Y,Z,W categories
To make it evident the criteria to choose the list of categories.

product-catalog/test/productcatalog/services/ProductProjectionServiceTest.java
```diff
@@ -0,0 +1,55 @@
...
+        final String otherSku = "M0E20000000DSBA";
+
+        assertThat(service.findVariantBySku(product, masterSku).get().getSku()).contains(masterSku);
+        assertThat(service.findVariantBySku(product, otherSku).get().getSku()).contains(otherSku);
+        assertThat(service.findVariantBySku(product, "")).isEmpty();
+    }
+
+    @Test
+    public void testGetSuggestions() {
+        final PlayJavaSphereClient sphereClient = new PlayJavaSphereClient() {
```
lauraluiz: I'd move this implementation to a method, giving it a meaningful name about what kind of sphere client it is trying to implement. We should leave the tests clean and focused just on what it is important.

common/app/common/pages/RatingData.java
lauraluiz: That's great, thanks @trangng @CT-Szilvia !

product-catalog/app/productcatalog/services/ProductProjectionService.java
lauraluiz: We can also use `IntStream.range(0, n)` ;)
Although Java devs might prefer the traditional way, I know.

product-catalog/app/productcatalog/pages/ProductDetailPageContent.java
lauraluiz: "Breadcrumb" is a template element, which happens to be a group of links shaped as a tree. At least I'm never referring to it as "Breadcrumb links".

app/inject/ShippingMethodProvider.java
lauraluiz: Is it now?

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
peter-gerhard: We need information from the product as well as from the specific variant. From the variant I can not get the product, so I fetch the product and get the specific variant by sku. Te later doesn't involve communication with sphere, so i guess it's no problem.

product-catalog/app/productcatalog/controllers/ProductCatalogController.java
peter-gerhard: I agree with renaming, but i disagree with the proposed names. The suggestions aren't necessarily   similar, in fact currently they are not at all. I suppose that the random picking of products is just temporary and will be replaced.

```getProductsFromCategories```/```getProductsByCategories``` ?

File: app/inject/ShippingMethodProvider.java (hunk 30 above)
Comment (peter-gerhard) [inline app/inject/ShippingMethodProvider.java]: @schleichardt Is it?
@lauraluiz I'm not sure if we can use this
>cartId - String - Required
The ID of the cart with a shipping address set.

When a fresh user enters the page is her/his shipping address set?

File: common/app/common/contexts/UserContext.java (hunk 31 above)
Comment (peter-gerhard) [inline common/app/common/contexts/UserContext.java]: I hope to get rid of one, they are not different concepts

File: common/app/common/pages/ImageData.java (hunk 32 above)
Comment (peter-gerhard) [inline common/app/common/pages/ImageData.java]: There was logic in the template
>{{#if thumbImage}}src="{{thumbImage}}"{{/if}} src="{{placeholderImage}}"

If I return an empty String this will not work.
I was satisfying that because the placeholderImage will not be in ImageData in the future but in static data. But I don't like that there is this logic in template.

File: product-catalog/app/productcatalog/controllers/ProductCatalogController.java (hunk 26 above)
Comment (butenkor) [inline product-catalog/app/productcatalog/controllers/ProductCatalogController.java]: If it is single API call, fine. I got confused that you used ```productService``` twice. I am used to use services for communication with DB or API only and util classes for stuff which does not need to be mocked.

File: product-catalog/app/productcatalog/controllers/ProductCatalogController.java (hunk 27 above)
Comment (lauraluiz) [inline product-catalog/app/productcatalog/controllers/ProductCatalogController.java]: `getRelatedProducts`? "related" doesn't necessary mean it is similar, but in the end in here it should appear something in particular for that product.

File: app/inject/ShippingMethodProvider.java (hunk 30 above)
Comment (schleichardt) [inline app/inject/ShippingMethodProvider.java]: in M16: http://sphereio.github.io/sphere-jvm-sdk/javadoc/master/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByCartGet.html

File: common/app/common/pages/ImageData.java (hunk 32 above)
Comment (lauraluiz) [inline common/app/common/pages/ImageData.java]: We have to take a look into it, because in Handlebars.js it says:
```
You can use the if helper to conditionally render a block. If its argument returns false, undefined, null, "", 0, or [], Handlebars will not render the block.
```
http://handlebarsjs.com/builtin_helpers.html

I'll check in Handlebars.java, maybe it's a change of behaviour.

lauraluiz: It shouldn't be behaving like that since 2014. According to this test still works as it should:

https://github.com/jknack/handlebars.java/blob/master/handlebars/src/test/java/com/github/jknack/handlebars/Issue109.java#L21

We can check it together.

PR #212, opened 2015-12-03
Several improvements for the catalog

lauraluiz:
Just refactoring and more.
Please @schleichardt review but I'll merge it right away. Any comments I will apply them next week.


schleichardt:
Ok, so I will review it on Tuesday, if you don't mind.

lauraluiz:
Not at all :)
And sorry for the mixing of topics, I was in the plane moving things around until I was satisfied with the whole picture.

PR #233, opened 2016-01-05
Enable country and language in configuration as override, closes #220

### LINKED ISSUE #220: Use project country/languages with override from config
So allow to override this values in configuration, but primary get them from project endpoint.

lauraluiz:
@schleichardt please review.
Now countries and languages are by default coming from the project info, unless defined in configuration file otherwise. This allows the full control with the CTP project but still provides the flexibility of defining your own.
This affected Handlebars, which now tries to load the languages coming from the project/config.
Also affects the checkout, which loads the countries from project/config.
And of course the language/country list, which is filled with info coming from project/config.

### DIFF
--- a/common/app/common/contexts/ProjectContext.java
+++ b/common/app/common/contexts/ProjectContext.java
@@ -5,11 +5,15 @@ import com.neovisionaries.i18n.CountryCode;
 import java.util.List;
 import java.util.Locale;
 
+import static io.sphere.sdk.utils.IterableUtils.requireNonEmpty;
+
 public class ProjectContext {
     private final List<Locale> languages;
     private final List<CountryCode> countries;
 
     private ProjectContext(final List<Locale> languages, final List<CountryCode> countries) {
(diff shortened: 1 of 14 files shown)


common/app/common/contexts/ProjectContext.java
```diff
@@ -22,6 +26,14 @@ private ProjectContext(final List<Locale> languages, final List<CountryCode> cou
         return countries;
     }
 
+    public Locale defaultLanguage() {
+        return languages.stream().findFirst().get();
```
schleichardt: `languages.get(0)` would be sufficient

schleichardt:
:+1:

common/app/common/contexts/ProjectContext.java
lauraluiz: I can't it hurts my eyes :P
I put it like this because I prefer how it looks like and theoretically it should do the same without performance loss.

PR #237, opened 2016-01-07
I18n messages, closes #236

### LINKED ISSUE #236: Use translations from YAML files only

lauraluiz:
@schleichardt please review.
Unfortunately it is a bit big although it is exclusively a single feature.
- Interface I18nResolver (which is injected and replaces Messages from Play) with implementations YamlI18nResolver and CompositeI18nResolver (that allows to override the messages resolvers as we do with the templates).

That's about it :)


common/app/common/i18n/CompositeI18nResolver.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public final class CompositeI18nResolver extends Base implements I18nResolver {
+    private final List<I18nResolver> i18nResolvers;
+
+    private CompositeI18nResolver(final List<I18nResolver> i18nResolvers) {
+        this.i18nResolvers = i18nResolvers;
+    }
+
+    @Override
+    public Optional<String> get(final String bundle, final String key, final Locale locale, final Object... args) {
```
schleichardt: I think bundle, key and args belong stronger together so I would use locale as first parameter.

conf/application.conf
```diff
@@ -17,10 +17,13 @@ play.modules.enabled += "inject.ApplicationProductionModule"
 # The application configuration
 # ~~~~~
 application.settingsWidget.enabled = true
-
-# overridden configuration from CTP project, default first
-#application.countries = ["DE", "US", "GB", "AT", "CH"]
-#application.languages = ["en", "de", "es"]
+#application.countries = ["DE", "US", "GB", "AT", "CH"] # uncomment to override countries from CTP project, default first
+#application.i18n.languages = ["en", "de", "es"] # uncomment to override languages from CTP project, default first
```
schleichardt: Does DE-de work?

schleichardt:
:+1:

common/app/common/i18n/CompositeI18nResolver.java
lauraluiz: Yes. I put args at the last moment so I just could put it in the end because of the ellipsis. Also I was still rethinking about making a class that already contains the locale... Well, I'll start changing the order.

conf/application.conf
lauraluiz: Yes, of course. It is just reading and parsing them into locales. Then the usage you do with those locales it's up to each particular situation. Exactly the same would happen if those locales are coming from CTP.

Issue #232, opened 2015-12-30
Improve error message display


lauraluiz:
Currently shown as JSON data

PR #246, opened 2016-01-12
Group modules, closes #231

### LINKED ISSUE #231: Group purchase module classes

lauraluiz:
@schleichardt would be great if you could pull this branch and review the new structure (in Github it will be impossible).


shopping-cart/app/shoppingcart/ErrorBean.java
```diff
@@ -1,4 +1,4 @@
-package purchase;
```
schleichardt: I think ErrorBean and ErrorsBean could be in the global common so that error handling is the same for all modules.

schleichardt:
:+1: looks good, it is much more cleaner now, thanks

shopping-cart/app/shoppingcart/ErrorBean.java
lauraluiz: Yep, that's why I left them in the root package, because I plan on moving them to a common place https://github.com/sphereio/sphere-sunrise/issues/232 :)

PR #250, opened 2016-01-14
Country and language switcher, closes #219, closes #155

### LINKED ISSUE #155: Enable changing user context
Currently the data is hardcoded and static.

### LINKED ISSUE #219: Change country/language
Currently languages and countries are displayed, allow to actually apply the changes.

lauraluiz:
@schleichardt please review.

- Implemented basic language switcher that redirects to home page (best would be to keep the user in the same page but let's wait in case we receive a little help from the platform).
- Implemented country switcher that saves country in session and redirects to home page (since you change the country and the whole shop might change, I think this is the best thing to do).
- Disabled selecting country for shipping address in checkout process, since you already selected the country. Allowing to select again the country in the middle of the checkout is too much effort and makes not much sense. The country selection is still available for billing address though, since it might differ from shipping and we are flexible there.

Missing to update currency in existing cart (also hoping to get some help here).


app/controllers/ApplicationController.java
```diff
@@ -25,16 +30,66 @@ public ApplicationController(final Injector injector, final SetupController setu
...
+            return language;
+        }
+
+        public void setLanguage(final String language) {
+            this.language = language;
+        }
+    }
+
+    public static class CountryFormData extends Base {
+        @Constraints.Required
```
schleichardt: Should we also validate that it has the right structure to be country? So what if I submit "1"?

app/inject/I18nResolverProvider.java
```diff
@@ -32,7 +32,7 @@ public I18nResolverProvider(final Configuration configuration, final ProjectCont
 
     @Override
     public I18nResolver get() {
-        final List<Locale> locales = projectContext.languages();
+        final List<Locale> locales = projectContext.locales();
```
schleichardt: This makes it harder to understand. So what is locales?
So if you have language "de" and have country "DE" you still need locale "de-DE" for formatting prices right. But you want to search for language "de".

app/controllers/ApplicationController.java
lauraluiz: It will be saved "1", but when Sunrise tries to use it, it will be converted into `CountryCode.UNDEFINED`, which will be discarded eventually and used the project's default country instead. I think it's not that important evaluating it here with regex or similar, when the evaluation will happen later anyway. The same applies to languages.

common/app/common/controllers/SunriseController.java
```diff
@@ -133,8 +133,38 @@ protected RequestContext requestContext() {
...
+        acceptedLocales.addAll(request.acceptLanguages().stream()
+                .map(lang -> Locale.forLanguageTag(lang.code()))
+                .collect(toList()));
+        acceptedLocales.addAll(projectContext.locales());
+        return acceptedLocales;
+    }
+
     @Nullable
-    public static String getCsrfToken(final Http.Session session) {
+    private static String getCsrfToken(final Http.Session session) {
```
schleichardt: Is this method still used? If no, it should be removed.

app/inject/I18nResolverProvider.java
lauraluiz: We have the same in `UserContext`. What you receive is a list of locales, not of languages. A locale can have regional variations while the language doesn't. I think therefore is more appropriate to call it locales instead of languages. Also, for me `languages` would be more a String that contains the meaning of a language in any form, but in this case is an instance of `Locale`.

common/app/common/models/LocationSelector.java
```diff
@@ -48,10 +49,11 @@ public void setCountry(final List<SelectableData> country) {
                     }
                     return selector;
                 }).collect(toList());
+        return (countrySelector.size() > 1) ? countrySelector : emptyList();
```
schleichardt: For some weird lit implementation size is expensive, so I would prefer `!countrySelector.isEmpty()`, also it does not require a magic number.

common/app/common/controllers/SunriseController.java
lauraluiz: It is used. This fills the meta information. Why wouldn't it be used? I put it as private because actually it is called from the same `SunriseController`.

common/app/common/models/LocationSelector.java
lauraluiz: `countrySelector.isEmpty()` == `countrySelector.size() > 0`, not `countrySelector.size() > 1` ;)

conf/application.conf
```diff
@@ -18,6 +18,7 @@ play.modules.enabled += "inject.ApplicationProductionModule"
 # ~~~~~
 application.settingsWidget.enabled = true
 #application.countries = ["DE", "US", "GB", "AT", "CH"] # uncomment to override countries from CTP project, default first
+application.currencies = ["EUR"] # uncomment to override currencies from CTP project, default first
```
schleichardt: It is not commented. Is this really right?

shopping-cart/app/shoppingcart/common/CartController.java
```diff
@@ -26,18 +28,43 @@ public CartController(final ControllerDependency controllerDependency) {
...
     }
 
+    private F.Promise<Cart> createCart(final UserContext userContext) {
+        final CartDraft cartDraft = CartDraft.of(userContext.currency()).withCountry(userContext.country());
+        return sphere().execute(CartCreateCommand.of(cartDraft))
+                .flatMap(cart -> updateCartCountry(cart, userContext.country()));
+    }
+
+    private F.Promise<Cart> fetchCart(final String cartId, final UserContext userContext) {
+        final CartByIdGet query = CartByIdGet.of(cartId).withExpansionPaths(m -> m.shippingInfo().shippingMethod());
```
schleichardt: Hmm, fetch cart would for me imply that you don't change the cart.

shopping-cart/app/shoppingcart/common/CartController.java
```diff
@@ -26,18 +28,43 @@ public CartController(final ControllerDependency controllerDependency) {
...
                 });
     }
 
+    private F.Promise<Cart> createCart(final UserContext userContext) {
+        final CartDraft cartDraft = CartDraft.of(userContext.currency()).withCountry(userContext.country());
+        return sphere().execute(CartCreateCommand.of(cartDraft))
+                .flatMap(cart -> updateCartCountry(cart, userContext.country()));
+    }
+
+    private F.Promise<Cart> fetchCart(final String cartId, final UserContext userContext) {
```
schleichardt: Hmm, the name "fetchCart" would for me imply that you don't change the cart.

conf/application.conf
lauraluiz: Yes, because I can't get the currencies from the Project (missing field in `Project` class), so I have no other option than to use them from the config file.

PR #253, opened 2016-01-19
Mini cart content, closes #221

### LINKED ISSUE #221: Display mini cart items and prices

lauraluiz:
@schleichardt please review, it also includes the refactor of cart related classes.


common/app/common/models/AttributeBean.java
```diff
@@ -0,0 +1,62 @@
+package common.models;
+
+import common.contexts.UserContext;
+import io.sphere.sdk.models.Base;
+import io.sphere.sdk.products.attributes.Attribute;
+import io.sphere.sdk.products.attributes.AttributeAccess;
+import io.sphere.sdk.products.attributes.AttributeExtraction;
+import io.sphere.sdk.producttypes.MetaProductType;
+
+public class AttributeBean extends Base {
```
schleichardt: Maybe ProductAttributeBean? A lot of things have attributes.

common/app/common/models/AttributeBean.java
```diff
@@ -0,0 +1,62 @@
...
+    private String name;
+    private String key;
+    private String value;
+
+    public AttributeBean() {
+    }
+
+    public AttributeBean(final Attribute attribute, final MetaProductType metaProductType, final UserContext userContext) {
```
schleichardt: We later need to adjust this since I want to deprecate MetaProductType. But for this pr ok.

common/app/common/models/AttributeBean.java
```diff
@@ -0,0 +1,62 @@
...
+        this.name = metaProductType
+                .findAttribute(attribute.getName())
+                .map(def -> def.getLabel().find(userContext.locales()).orElse(""))
+                .orElse(null);
```
schleichardt: It is weird that one time `orElse` gets `null` as argument and the other time `null`.

common/app/common/models/AttributeBean.java
```diff
@@ -0,0 +1,62 @@
...
+    public void setValue(final String value) {
+        this.value = value;
+    }
+
+    protected static String formatValue(final MetaProductType metaProductType, final Attribute attribute,
+                                        final UserContext userContext) {
+        final AttributeExtraction<String> attributeExtraction = AttributeExtraction.of(metaProductType, attribute);
+        return attributeExtraction
+                .ifIs(AttributeAccess.ofLocalizedString(), v -> v.find(userContext.locales()).orElse(""))
+                .ifIs(AttributeAccess.ofLocalizedEnumValue(), v -> v.getLabel().find(userContext.locales()).orElse(""))
```
schleichardt: Wouldn't be http://sphereio.github.io/sphere-jvm-sdk/javadoc/master/io/sphere/sdk/models/LocalizedString.html#getTranslation-java.lang.Iterable- more suitable (also on the other places here)

common/app/common/models/ProductVariantBean.java
```diff
@@ -6,91 +6,53 @@
...
     }
 
-    public void setDescription(final String description) {
-        this.description = description;
+    private void fillPrice(final Price price, final Locale locale) {
+        final MoneyContext moneyContext = MoneyContext.of(price.getValue().getCurrency(), locale);
+        final MonetaryAmount currentPrice = calculateFinalPrice(price);
+        final boolean hasDiscount = currentPrice.isLessThan(price.getValue());
+        if (hasDiscount) {
+            this.priceOld = moneyContext.formatOrNull(price);
```
schleichardt: If the old price is null, should then the `hasDiscount` flag set to false?

common/app/shoppingcart/CartSessionUtils.java
```diff
@@ -11,21 +16,22 @@
...
+        return Optional.ofNullable(session.get(CartSessionKeys.MINI_CART))
+                .map(miniCartAsJson -> SphereJsonUtils.readObject(miniCartAsJson, MiniCart.class))
+                .orElse(new MiniCart());
     }
 
-    public static void overwriteCartSessionData(final Cart cart, final Session session) {
-        final long itemCount = cart.getLineItems().stream().mapToLong(LineItem::getQuantity).sum();
+    public static void overwriteCartSessionData(final Cart cart, final Session session, final UserContext userContext,
+                                                final ReverseRouter reverseRouter) {
+        final JsonNode miniCartAsJson = SphereJsonUtils.toJsonNode(new MiniCart(cart, userContext, reverseRouter));
```
schleichardt: If the routes of the shop change, will then the minicart point to false urls?

shopping-cart/app/shoppingcart/common/ExtendedLineItemBean.java
```diff
@@ -0,0 +1,36 @@
...
+import common.models.AttributeBean;
+import common.models.LineItemBean;
+import common.models.ProductDataConfig;
+import io.sphere.sdk.carts.LineItem;
+
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+public class ExtendedLineItemBean extends LineItemBean {
```
schleichardt: Why can't this be in the LineItemBean itself?

schleichardt: I am still confused about the purpose and structure of LineItemBean and ExtendedLineItemBean.
Can you document them and/or give them better names?

common/app/common/models/AttributeBean.java
lauraluiz: Yes. For some reason I was convinced before it was named like this.

common/app/common/models/AttributeBean.java
lauraluiz: Probably it comes from old code, since I only extracted existing code. I will change it to empty string, since it can be save NPE on other template engines.

common/app/common/models/AttributeBean.java
lauraluiz: I started using this method, but I rollbacked for these reasons:
1) it is unlikely that you have `en-UK` and `en` defined in your project, you would usually have `en-UK` and `en-US` in any case, and this is not helpful for the method.
2) we already use several fallbacks to make sure we get some language (user-agent preferred languages, project default language)

Of course these are not reasons to stop using this method, because these are just arguments to evidence that is not a great help in this case. The real reason is that it will pollute the code with `Optional.ofNullable(v.getLabel().getTranslation(userContext.locales())).orElse("")` which is not so nice anymore and again we require some util class to support it (which anyway is not a bad idea, as it is easier to change the way of getting the translations).

Anyway, not a topic for this PR, but I'll probably try some idea in future PR.

common/app/common/models/ProductVariantBean.java
lauraluiz: I'm not sure which flag do you mean

shopping-cart/app/shoppingcart/common/ExtendedLineItemBean.java
lauraluiz: Extended has more info, which is not required for a simplified version of the cart (i.e. mini-cart). Right now that means it additionally has attributes, but it can have more info in the future (i.e. availability, shipping info...). I cannot figure out any other name. I'm also not sure what to document, I can only say it has more info, but I guess that's evident by the name. Can you give suggestions?

common/app/shoppingcart/CartSessionUtils.java
lauraluiz: That's a complex thing to solve. So far this version works well enough, I'll push solving this issue post-release: https://github.com/sphereio/sphere-sunrise/issues/254

shopping-cart/app/shoppingcart/common/ExtendedLineItemBean.java
lauraluiz: Okay so I renamed the simplified version as "MiniCartLineItem" and the extended version as "CartLineItem". I think it makes sense that a cart object extends a mini cart object, since they are basically the same but with more/less info.

PR #265, opened 2016-01-22
More renaming, closes #248

### LINKED ISSUE #248: Rebrand to commercetools
We should focus on the brand commercetools and remove all unnecessary mentions of SPHERE.IO

lauraluiz:
And I used the wrong branch :)


schleichardt:
@lauraluiz was it intentionally to delete the country stuff?

lauraluiz:
Yes, I always saw it and I thought "I'll remove it in the next PR to avoid polluting". Now I thought I was just pushing to master and I took advantage and removed it... but I wasn't :)

The reason is: we are not using those classes, we deal with countries differently.

PR #267, opened 2016-01-25
Optimization

lauraluiz:
Some optimizations:
- Allow to enable/disable CTP metrics
- Remove `color` attribute, since data is by default with only one color
- Clean up some code that might slightly affect performance


lauraluiz:
@schleichardt can you take a look?

schleichardt:
:+1:

PR #271, opened 2016-01-28
Review Handlebars

lauraluiz:
- Update to latest handlebars version 4.0.3.
- Remove cache option: now it is initialized with a TemplateCache and it does not seem to be affected by Play's hot-reloading.
- Remove fallback context option: it only supported JSON and was a bit hacky, plus I cannot find any real use to be enabled as configuration.

### DIFF
--- a/common/app/common/templates/HandlebarsTemplateService.java
+++ b/common/app/common/templates/HandlebarsTemplateService.java
@@ -1,40 +1,31 @@
 package common.templates;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.jknack.handlebars.Context;
 import com.github.jknack.handlebars.Handlebars;
 import com.github.jknack.handlebars.Template;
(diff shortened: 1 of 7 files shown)


common/app/common/templates/HandlebarsTemplateService.java
```diff
@@ -1,40 +1,31 @@
...
         this.handlebars = handlebars;
-        this.fallbackContexts = fallbackContexts;
-        this.cacheEnabled = cacheEnabled;
     }
 
     @Override
     public String render(final String templateName, final PageData pageData, final List<Locale> locales) {
         final Template template = compileTemplate(templateName);
-        final Context context = buildContext(pageData, templateName);
+        final Context context = Context.newContext(pageData);;
```
schleichardt: duplicate semicolon could be removed

schleichardt:
:+1:

PR #285, opened 2016-02-09
Fix and test copier to create also missing folders, fixes #281

### LINKED ISSUE #281: Fix sbt copy command for unmanaged resources

lauraluiz:


app/tasks/WebjarsFilesCopier.java
```diff
@@ -0,0 +1,56 @@
...
+        });
+    }
+
+    private static void copyFile(final File origFile, final Path destPath) {
+        LOGGER.debug(origFile + " => " + destPath);
+        try {
+            FileUtils.copyFileToDirectory(origFile, destPath.toFile());
+            LOGGER.info("Successfully created \"{}\"", destPath);
+        } catch (FileAlreadyExistsException e) {
+            LOGGER.error("File \"{}\" already exists, please delete it first if you want to replace it", destPath);
```
schleichardt: you could also pass the exception to the logger

lauraluiz: Thought about it, but the people that will use this don't care about the code details (the exception is not giving info of their interest) and it will hide the message, which I want it to be easy to spot to ready it. That's why I thought against it.

PR #304, opened 2016-02-18
Include default configuration in Heroku, closes #297

### LINKED ISSUE #297: Add default values to app.json for Heroku

lauraluiz:
Also re-structure the configuration so it looks nicer (@schleichardt I think I managed to make it nice)

### DIFF
--- a/app.json
+++ b/app.json
@@ -82,15 +82,15 @@
     },
 
     "COUNTRIES": {
-      "description": "OPTIONAL! LEAVE EMPTY IF YOU WANT THE APPLICATION TO USE THE COUNTRIES DEFINED IN YOUR PROJECT. If you want to replace them, enter the countries as ISO country codes (e.g. DE) separated by commas. Default country first.",
+      "description": "OPTIONAL! LEAVE EMPTY IF YOU WANT THE APPLICATION TO USE THE COUNTRIES DEFINED IN YOUR COMMERCETOOLS PROJECT. If you want to replace them, enter the countries as ISO country codes (e.g. DE) separated by commas. Default country first.",
       "required": false
     },
     "CURRENCIES": {
-      "description": "OPTIONAL! LEAVE EMPTY IF YOU WANT THE APPLICATION TO USE THE CURRENCIES DEFINED IN YOUR PROJECT. If you want to replace them, enter the currencies as ISO currency codes (e.g. EUR) separated by commas. Default currency first.",
+      "description": "OPTIONAL! LEAVE EMPTY IF YOU WANT THE APPLICATION TO USE THE CURRENCIES DEFINED IN YOUR COMMERCETOOLS PROJECT. If you want to replace them, enter the currencies as ISO currency codes (e.g. EUR) separated by commas. Default currency first.",
       "required": false
     },
     "LANGUAGES": {
-      "description": "OPTIONAL! LEAVE EMPTY IF YOU WANT THE APPLICATION TO USE THE LANGUAGES DEFINED IN YOUR PROJECT. If you want to replace them, enter the languages as language tags (e.g. en-UK) separated by commas. Default language first.",
+      "description": "OPTIONAL! LEAVE EMPTY IF YOU WANT THE APPLICATION TO USE THE LANGUAGES DEFINED IN YOUR COMMERCETOOLS PROJECT. If you want to replace them, enter the languages as language tags (e.g. en-UK) separated by commas. Default language first.",
       "required": false
     },
 
--- a/app.json
+++ b/app.json
@@ -82,21 +82,21 @@
     },
 
     "COUNTRIES": {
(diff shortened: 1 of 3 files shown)


conf/application.conf
```diff
@@ -1,49 +1,124 @@
...
+
+# PRODUCTS
 # ~~~~~
-# The secret key is used to secure cryptographics functions.
-# If you deploy your application to several instances be sure to use the same key!
-play.crypto.secret="changeme"
-play.crypto.secret=${?APPLICATION_SECRET}
 
-# Injection configuration
+# Product attribute list that is displayed for a specific product variant (this does not allow to select different
```
schleichardt: how about `Product attribute field name list` to make clear what the value is?

conf/application.conf
```diff
@@ -1,49 +1,124 @@
...
+# product variants, please check soft/hard selectable attributes for that purpose). Enter a list of the attribute keys.
+productData.displayedAttributes = ["colorFreeDefinition", "size", "designer"]
+productData.displayedAttributes = ${?DISPLAYED_ATTRIBUTES}
+
+# Product attribute list that allows to select a clearly different variant of your product, which requires to load the
+# new product information associated with it (e.g. color). Enter a list of the attribute keys.
+productData.hardSelectableAttributes = ["color"]
+productData.hardSelectableAttributes = ${?HARD_SELECTABLE_ATTRIBUTES}
+
+# Product attribute list that allows to select a specific variant, but does not require to change the displayed product
```
schleichardt: I don't like the term change here since it could sound like updating a product, how about `reload the displayed product for the selected variant`?

conf/application.conf
```diff
@@ -1,49 +1,124 @@
...
+  {"message":"form.personTitles.dr", "value":"dr"}
+]
+
+checkout.customerServiceNumber = "+49 899982996-0"
+
+
+# LOCALIZATION
+# ~~~~~
+
+# LEAVE COMMENTED IF YOU WANT THE APPLICATION TO USE THE COUNTRIES DEFINED IN YOUR PROJECT!
```
schleichardt: I think project is to generic, how about
`LEAVE COMMENTED IF YOU WANT THE APPLICATION TO USE THE COUNTRIES DEFINED IN YOUR COMMERCETOOLS PROJECT!`?

conf/application.conf
lauraluiz: You are not displaying the product attribute field name, but the product attribute (which consists of name and value). You have to enter the product attribute keys in this configuration, but that's explained later.

conf/application.conf
lauraluiz: You think too much from a backend perspective :P
But I'll put reload, as I mention "load" in the hard-attributes, so it would fit.

conf/application.conf
```diff
@@ -1,49 +1,124 @@
 # This is the main configuration file for the application.
```
schleichardt: could you also link https://github.com/typesafehub/config in the top, so that people new to this configuration mechanism can look up the syntax and features?

lauraluiz: Done

schleichardt:
I like the application.conf and its documentation.

conf/application.conf
lauraluiz: Yep. I have also added it to app.json (Heroku)

conf/application.conf
schleichardt: sorry, but the attribute element has no key, it is called 'name'

lauraluiz: Then I will simply replace `key` by `name`

lauraluiz: So I changed it to `name`, which is the official name we use everywhere. Maybe to avoid confusion with `label` I may add some clarification if I see somebody has trouble with it, but now I'm going to see how it goes like this.
I don't want to use `attribute field name` because to me it sounds too much like `custom field name` and I think it has the same chances to confuse than to help.

app.json
```diff
@@ -16,47 +16,20 @@
...
 
     "HARD_SELECTABLE_ATTRIBUTES": {
-      "description": "Product attribute list that allows to select a clearly different variant of your product, which requires to load the new product information associated with it (e.g. color). Enter the attribute keys separated by commas.",
+      "description": "List here those product attributes that enable selecting a different product variant. In particular, those attributes that require re-loading the page to display the selected variant, e.g. if you select a different color you may like to show different pictures, description, etc. Enter the attribute names separated by commas.",
       "required" : false
     },
 
     "SOFT_SELECTABLE_ATTRIBUTES": {
-      "description": "Product attribute list that allows to select a specific variant, but does not require to change the displayed product information (e.g. size). Enter the attribute keys separated by commas.",
+      "description": "List here those product attributes that enable selecting a different product variant. In particular, those attributes that DO NOT require re-loading the page to display the selected variant, e.g. size. Enter the attribute names separated by commas.",
```
lauraluiz: One more try @stmeissner

PR #307, opened 2016-02-22
Sign in/up/out, closes #305

### LINKED ISSUE #305: Enable customer log in

lauraluiz:
@schleichardt please review.

Among the implementation of the sign in/up/out itself, there are some other minor improvements I do on the classes I have to touch. This way I make sure the code gets better even though I don't have an official task for that. This time it was ReverseRouter, routes and CartSession's turn.

I hope it doesn't get tedious to review despite of that.


my-account/app/myaccount/login/LogInPageController.java
```diff
@@ -0,0 +1,165 @@
...
+            return F.Promise.pure(handleSignUpFormErrors(form, pageContent, userContext));
+        } else {
+            return signUp(form)
+                    .map(signInResult -> handleSuccessfulSignIn(signInResult, userContext))
+                    .recover(throwable -> handleExistingCustomerError(throwable, pageContent, userContext));
+        }
+    }
+
+    public Result processLogOut(final String languageTag) {
+        CustomerSessionUtils.removeCustomer(session());
```
schleichardt: should we add here sth. more generic? What if another page part needs also to clean up the session?
Can we just clear the whole session?

my-account/app/myaccount/login/LogInPageController.java
```diff
@@ -0,0 +1,165 @@
...
+        CartSessionUtils.removeCart(session());
+        return redirect(reverseRouter().showHome(languageTag));
+    }
+
+    private F.Promise<CustomerSignInResult> logIn(final Form<LogInFormData> form) {
```
schleichardt: I would use as parameter just the LogInFormData and the caller does the get.
When I first read this method I thought you did not check the form for errors.

my-account/app/myaccount/login/SignUpFormData.java
```diff
@@ -0,0 +1,88 @@
+package myaccount.login;
+
+import io.sphere.sdk.models.Base;
+import play.data.validation.Constraints;
+
+public class SignUpFormData extends Base {
+    @Constraints.Required
```
schleichardt: everybody needs a title to sign up?

I think the form is screwed up with salutation.

schleichardt:
apart from my comments :+1:

my-account/app/myaccount/login/LogInPageController.java
lauraluiz: It's difficult to say. In the particular case of cart/customer there's a clear dependency of customer -> cart, but cart -> customer is not true (even though I cannot really think of an use case). And different information might be even less related. Anyway, I see this topic as a bigger task for the future when we have at least one more session case (e.g. wishlist would be a good case). I've created a task for it: https://github.com/sphereio/commercetools-sunrise-java/issues/309

my-account/app/myaccount/login/SignUpFormData.java
lauraluiz: It is a dropdown where you have a pre-selected value, so I put `Required` because it's practically true (i.e. it's not possible via the UI to put it optional). Also usually in shops they need to know the title to write all the emails and letters, so it's usually mandatory because of that.

But what do you mean with "the form is screwed up with salutation"?

PR #362, opened 2016-04-21
Enable demo info modal

lauraluiz:


conf/i18n/en/modal-announce.yaml
```diff
@@ -0,0 +1,5 @@
+infoModal:
+  title: Dear guest,
+  body: <p>You are visiting our current <b>commercetools</b> demo shop <b>SUNRISE</b>.</p><p>If you are interested in a personal demonstration, please contact our sales team via email <b>sales@commercetools.de</b> or by phone <b>+49 (89) 99 82 996-28</b>.</p>
```
schleichardt: Shouldn't we switch to the com domain email address?

PR #363, opened 2016-04-22
My orders, close #331

### LINKED ISSUE #331: My Account: My Orders
http://sphereio.github.io/commercetools-sunrise-theme/site/en/my-account-my-orders.html

lauraluiz:
@schleichardt please review. Not only I've implemented my orders, but I've fixed and improved the payment scenario (after an issue with the order, where the payment was not expanded).


my-account/app/myaccount/login/LogInPageController.java
```diff
@@ -66,7 +66,7 @@ public LogInPageController(final ControllerDependency controllerDependency, fina
         } else {
             return logIn(logInForm.get())
                     .thenComposeAsync(signInResult -> handleSuccessfulSignIn(signInResult, userContext), HttpExecution.defaultContext())
-                    .exceptionally(throwable -> handleInvalidCredentialsError(throwable, logInForm, userContext).toCompletableFuture().join()); // TODO move to async
+                    .exceptionally(throwable -> handleInvalidCredentialsError(throwable, logInForm, userContext).toCompletableFuture().join()); // TODO move to async and send HTTP context!
```
lauraluiz: https://github.com/sphereio/commercetools-sunrise-java/issues/365

my-account/app/myaccount/orders/MyOrdersPageController.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.CompletionStage;
+
+import static java.util.Collections.emptyList;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static java.util.stream.Collectors.toList;
+
+@Singleton
+public class MyOrdersPageController extends MyAccountController {
+
+    public static final int PAGE_SIZE = 500; // TODO allow to configure and work with pages
```
lauraluiz: Theme issue

common/app/common/utils/CtpUtils.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class CtpUtils {
+
+    /**
+     * Transforms a CTP enum name (e.g. BALANCE_DUE) to lower CamelCase (e.g. balanceDue).
+     * @param ctpEnumName the CTP enum name to transform into lower CamelCase
+     * @return the lower CamelCase version of the CTP enum name
+     */
+    public static String enumToCamelCase(final String ctpEnumName) {
+        return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, ctpEnumName);
```
schleichardt: maybe this is sth. for the SDK, so it could be moved to "move-to-sdk"

common/app/common/utils/ArrayUtils.java
```diff
@@ -0,0 +1,30 @@
...
+     * @return the element from the array in that position, or the default value if there is no element
+     */
+    @Nullable
+    public static String getArrayElement(final String[] array, final int position, final String defaultValue) {
+        final boolean containsPosition = array.length > position;
+        return containsPosition ? array[position] : defaultValue;
+    }
+
+    public static <T> List<T> arrayToList(final T[] array) {
+        return asList(array);
```
schleichardt: I would prefer in that cases to directly use the JDK method instead of "arrayToList".

common/app/common/utils/FutureUtils.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Transforms a list of {@code CompletionStage} into a {@code CompletionStage} of a list,
+     * that will be completed once all the elements of the given list are completed.
+     * @param stageList list of {@code CompletionStage}
+     * @param <T> the element obtained from the list of {@code CompletionStage}
+     * @return the {@code CompletionStage} of a list of elements
+     */
+    public static <T> CompletionStage<List<T>> listOfFuturesToFutureOfList(final List<CompletionStage<T>> stageList) {
```
schleichardt: could be added here:
http://sphereio.github.io/sphere-jvm-sdk/apidocs/io/sphere/sdk/utils/CompletableFutureUtils.html

schleichardt: btw: I think you searched this method: http://sphereio.github.io/sphere-jvm-sdk/apidocs/io/sphere/sdk/utils/CompletableFutureUtils.html#recoverWith-java.util.concurrent.CompletionStage-java.util.function.Function-

common/app/myaccount/CustomerSessionKeys.java
```diff
@@ -1,10 +0,0 @@
-package myaccount;
-
-public final class CustomerSessionKeys {
```
schleichardt: I liked the separation of responsibilities for keys and the behaviour.

conf/i18n/en/my-account.yaml
```diff
@@ -0,0 +1,66 @@
+myAccount:
```
schleichardt: Is there an issue to add the translation of my-account for German?

schleichardt:
ok

common/app/common/utils/ArrayUtils.java
lauraluiz: It is just one way to convert an array to a list. It can change over time what do we understand as the best way, and when that happens I want to be able to change it cleanly. I have always trouble remembering which was the best way and then I have to look for it. If I have it as a method I don't have to worry about it anymore.

common/app/common/utils/CtpUtils.java
lauraluiz: 👍

common/app/common/utils/FutureUtils.java
lauraluiz: `recoverWith` shouldn't it be triggered only when an exception happened? Not sure what you mean with it.

schleichardt: You had the problem that recover on CompletionStage does not take a function which returns a CompletionStage but recoverThis does.

common/app/myaccount/CustomerSessionKeys.java
lauraluiz: It made sense if somebody outside the session utils is actually using them, but nobody is accessing them, so the separation is just creating more classes (3 more). That's why after creating the OrderSessionUtils and realizing the usage outside the class is null, I decided to merge them. Of course once Sunrise is a framework it will be likely used outside, that's why I would rather rename the constants to something that indicates the meaning. Therefore if somebody is not finding the method they are looking for in the *SessionUtils, they can simply access the session key in the same class, instead of guessing they will find them in another class.

conf/i18n/en/my-account.yaml
lauraluiz: Maybe that I'm not German ;)
Anyway, in here it makes no sense to provide the German version where nobody can properly edit it. Once it is in the theme, then I will just ask anybody to translate it. The changes I do in the app are just to make it work.

common/app/common/utils/FutureUtils.java
lauraluiz: Oh you mean this: https://github.com/sphereio/commercetools-sunrise-java/issues/365?

conf/i18n/en/my-account.yaml
schleichardt: With issue - I meant ticket. ;)

right, the translations are also in the theme.

move-to-sdk/app/io/sphere/sdk/utils/FutureUtils.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Transforms a list of {@code CompletionStage} into a {@code CompletionStage} of a list,
+     * that will be completed once all the elements of the given list are completed.
+     * @param stageList list of {@code CompletionStage}
+     * @param <T> the element obtained from the list of {@code CompletionStage}
+     * @return the {@code CompletionStage} of a list of elements
+     */
+    public static <T> CompletionStage<List<T>> listOfFuturesToFutureOfList(final List<CompletionStage<T>> stageList) {
```
schleichardt: I will add it to the SDK:

https://github.com/sphereio/sphere-jvm-sdk/issues/1124

lauraluiz: 👍

move-to-sdk/app/io/sphere/sdk/utils/FutureUtils.java
```diff
@@ -29,4 +32,30 @@ private FutureUtils() {
                         .map(CompletableFuture::join)
                         .collect(Collectors.<T>toList()));
     }
+
+    public static <T> CompletionStage<T> exceptionallyCompletedFuture(final Throwable throwable) {
```
lauraluiz: @schleichardt I saw after I created the method that you had the same in the utils. The name though (although shorter, and I'm pro shorter names) doesn't fit very well the code when reading.
Having into mind that the positive return looks like:
`return completedFuture(value);`
Then the negative return for me is:
`return exceptionallyCompletedFuture(throwable);`

With your method could look like:
`return failed(throwable);`
`return CompletableFutureUtils.failed(throwable);`
So I'm not that inclined to use yours, that's why I left mine.

schleichardt: I will create an alias, see https://github.com/sphereio/sphere-jvm-sdk/issues/1128

lauraluiz: Great, thanks 👍

PR #377, opened 2016-05-03
Refactor checkout, closes #338

### LINKED ISSUE #338: Show billing address even when copied from shipping
Currently it is empty in the confirmation page when billing is using the same shipping address.

lauraluiz:
@schleichardt please review the changes I did, because I pretend to use this as a guidance on how to develop on Sunrise. If you agree, then I'll apply it to catalog, which is the only one missing this structure.

As the code is too much, here I list the many improvements I've implemented, towards unifying code and reusing beans:

- Avoid having logic in the `xPageContent` classes, instead let the Controller handle how to deal with it.
- Don't let models know about Play Forms.
- Have logic in the beans exclusively regarding the related JVM SDK class they are trying to extend, but don't give it too many responsibilities (e.g. ShippingInfoBean should only know how to extract data from shippingInfo, not from cart).
- Unify class names (e.g. FormData, FormBean, PageController, PageContent, Bean...)
- Unify controller methods and structure, special focus on the ones that handle forms (shopping-cart, my-account).
- Controller methods split into something that can be functional and overridable once we have it as a framework.
- Easily extract errors and include them in the data to the template (even though in some cases it is not actually used, the theme needs to be adapted).
- Follow standard Play way to extract data from a form, even though the form might contain errors (with `field`, which is true that might be error prone when typing the name, but 1st it is not big deal as it is harmless and 2nd it is more understandable code). This way getting the data to fill back the form in case of error is independent from getting the real data to send to CTP, and only done in each case (no duplication).
- Separated beans used to display forms (e.g. AddressFormBean) and beans used to show info (e.g. AddressBean).

And other functional improvements on checkout:
- Used Payment classes to store payment info.
- Show shipping address for billing when billing empty (i.e. same as shipping).
- Hide newsletter checkbox, which confused people.
- Find shipping methods applying for that cart, instead of fetching them at the beginning.

Happy reviewing :)


common/app/common/errors/ErrorsBean.java
```diff
@@ -7,21 +7,28 @@
...
     public ErrorsBean() {
     }
 
     public ErrorsBean(final String errorMessage) {
-        this.globalErrors = Collections.singletonList(new ErrorBean(errorMessage));
+        this.globalErrors = singletonList(new ErrorBean(errorMessage));
     }
 
-    public ErrorsBean(final Form<?> filledForm) {
+    public ErrorsBean(final List<Form<?>> filledForms) {
```
schleichardt: Multiple forms, why?

common/app/common/models/SelectableBean.java
```diff
@@ -2,22 +2,23 @@
 
 import io.sphere.sdk.models.Base;
 
-public class SelectableData extends Base {
+public class SelectableBean extends Base {
```
schleichardt: Can this name be improved by telling what is selected? Like FormSelectionBean?

common/app/common/models/TitleFormFieldBean.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private static final String CONFIG_TITLE_OPTIONS = "form.titles";
+
+    private List<SelectableBean> list;
+
+    public TitleFormFieldBean() {
+    }
+
+    public TitleFormFieldBean(@Nullable final String selectedTitle, final UserContext userContext,
+                              final I18nResolver i18nResolver, final Configuration configuration) {
```
schleichardt: Can we have a constructor that does not depend on Configuration?

Configuration is cool, but hard to know what you exactly use from it.

I like this class https://github.com/sphereio/jvm-trainings/blob/master/tarent-web/src/main/java/io/sphere/demo/tarent/commercetools/CommercetoolsPlatformConfig.java since it is very declarative extracting values from properties. Maybe there is something doable with guice and Configuration?

common/app/common/utils/FormUtils.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.annotation.Nullable;
+
+public final class FormUtils {
+
+    private FormUtils() {
+    }
+
+    public static Address extractAddress(@Nullable final Form<?> form, final String suffix) {
+        if (form != null) {
+            final CountryCode country = countryOrUndefined(form.field("country" + suffix).value());
```
schleichardt: It is very repetitive, would this be better?

```java
final Function<String, String> fieldExtractor = fieldName -> form.field(fieldName + suffix).value();
return AddressBuilder.of(country)
  .title(fieldExtractor.apply("title"))
  .firstName(fieldExtractor.apply("firstName))
  //etc.
  .build();
```

common/app/common/utils/FormUtils.java
```diff
@@ -0,0 +1,39 @@
...
+                    .streetName(form.field("streetName" + suffix).value())
+                    .additionalStreetInfo(form.field("additionalStreetInfo" + suffix).value())
+                    .city(form.field("city" + suffix).value())
+                    .postalCode(form.field("postalCode" + suffix).value())
+                    .region(form.field("region" + suffix).value())
+                    .phone(form.field("phone" + suffix).value())
+                    .email(form.field("email" + suffix).value())
+                    .build();
+        } else {
+            return Address.of(CountryCode.UNDEFINED);
```
schleichardt: cool, did not know that `CountryCode.UNDEFINED` exists

common/app/shoppingcart/AddressBean.java
```diff
@@ -28,54 +25,26 @@ public AddressBean() {
...
     public AddressBean(@Nullable final Address address, final Locale locale) {
         if (address != null) {
-            try {
-                BeanUtils.copyProperties(this, address);
-            } catch (IllegalAccessException | InvocationTargetException e) {
-                throw new RuntimeException(e);
-            }
-            setCountryCode(address.getCountry().getAlpha2());
-            setCountry(address.getCountry().toLocale().getDisplayCountry(locale));
+            this.title = address.getTitle();
```
schleichardt: `this.` is superflous

common/app/shoppingcart/PaymentInfoBean.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public PaymentInfoBean(final @Nullable PaymentInfo paymentInfo, final UserContext userContext) {
+        if (paymentInfo != null) {
+            this.type = paymentInfo.getPayments().stream()
+                    .filter(ref -> ref.getObj() != null)
+                    .map(ref -> {
+                        final PaymentMethodInfo methodInfo = ref.getObj().getPaymentMethodInfo();
+                        return methodInfo.getName().find(userContext.locales()).orElse(methodInfo.getMethod());
+                    })
+                    .findFirst() // TODO Handle multiple payments
```
schleichardt: add this TODO better to an issue tracker

my-account/app/myaccount/login/LogInPageController.java
```diff
@@ -35,66 +36,65 @@
...
+    public CompletionStage<Result> processLogOut(final String languageTag) {
         CustomerSessionUtils.removeCustomer(session());
         CartSessionUtils.removeCart(session());
-        return redirect(reverseRouter().showHome(languageTag));
+        final Call call = reverseRouter().showHome(languageTag);
+        return completedFuture(redirect(call));
     }
 
-    private CompletionStage<CustomerSignInResult> logIn(final LogInFormData formData) {
+    protected CompletionStage<CustomerSignInResult> logIn(final LogInFormData formData) {
```
schleichardt: If you intend to let people override your methods, you should document them well.

shopping-cart/app/shoppingcart/checkout/shipping/CheckoutShippingPageController.java
```diff
@@ -0,0 +1,140 @@
...
+        if (throwable.getCause() instanceof SphereException) {
+            return getShippingMethods(session())
+                    .thenApplyAsync(shippingMethods -> {
+                        Logger.error("Unknown CTP error", throwable.getCause());
+                        final ErrorsBean errors = new ErrorsBean("Something went wrong, please try again"); // TODO get from i18n
+                        final CheckoutShippingPageContent pageContent = createPageContentWithShippingError(shippingForm, errors, shippingMethods);
+                        return badRequest(renderCheckoutShippingPage(cart, pageContent, userContext));
+                    }, HttpExecution.defaultContext());
+        }
+        throw new RuntimeException(throwable);
```
schleichardt: that is weird, the return type is a CompletionStage, so you could return a failed CompletableFuture.
By using this method I would not expect a plain Runtime exception with the wrapped error. Maybe you want to throw an IllegalArgumentException since the method is not supposed to handle non commercetools exceptions?

schleichardt:
+1 and you are already assigned

common/app/common/errors/ErrorsBean.java
lauraluiz: Because initially I was sending the billingAddress and shippingAddress together and showing all the errors. But there could be too many and I thought better to send them separately, so I don't use the list of forms anymore. I thought about leaving it for the future, as it doesn't have the easiest code, to avoid losing it for the future.

common/app/common/models/SelectableBean.java
lauraluiz: Good point. Initially it was something generic, but now I realized it's only used for forms, so the name can be adapted.

common/app/common/models/TitleFormFieldBean.java
lauraluiz: Yes, I also wanted to remove the Configuration from there, also because it has to be loaded every single time. But wanted to aimed for that later, so thanks for the inspiration, I will try to change it once I have clear the best way:
https://github.com/sphereio/commercetools-sunrise-java/issues/379

common/app/common/utils/FormUtils.java
lauraluiz: Yep, I can also enable this extractor in this utils class, so that it can be used elsewhere.

my-account/app/myaccount/login/LogInPageController.java
lauraluiz: Yes, this will be handled when converting Sunrise into a framework, step by step:
https://github.com/sphereio/commercetools-sunrise-java/issues/368

common/app/shoppingcart/PaymentInfoBean.java
lauraluiz: Actually I put in TODO all those issues that cannot be handled by the app alone, so in this case it needs support from the theme. I need to recollect all those and create the corresponding issues in theme.

common/app/shoppingcart/AddressBean.java
lauraluiz: I like to use `this` always for constructors, it allows me to see it clearly when the assignment happened. Just a little hint for me.

shopping-cart/app/shoppingcart/checkout/shipping/CheckoutShippingPageController.java
lauraluiz: Yep, will do.

lauraluiz: Ah I just realised this method is not even used, it was just a draft for the future handling of CTP errors. But I need your help with that because of `exceptionally`.

PR #388, opened 2016-05-12
Docker image deployment

svenmueller:
This PR enables travis to build and push Docker images to the public Docker Hub and replaces the TC deployment.

On each push to the remote github repository, a Docker image is build and pushed by travis. The following Docker image are provided:

- latest (for master branch) or branch name (any other branch)
- short git commit SHA (first 8 chars), e.g. `11be0178`
- tag containing the travis build number, e.g. `travis-17`

## Deployment to Production

To deploy to production, a new git commit tag needs to be created.

This will trigger a new Travis build as described before but will create an additional Docker image tagged with `production. The github tag can be created via command line or github UI ("Draft new Release". https://github.com/commercetools/sphere-carhartt/releases)

`git tag -a v1.0.1 -m "Minor text adjustments."`

A new Docker image with tag `production` is pulled and deployed to Docker Cloud automatically.

@sphereio/sunrise-team @sphereio/ops please review


anas-aso:
@svenmueller there seem to be some errors on Travis https://travis-ci.org/sphereio/commercetools-sunrise-java/builds/129736970

schleichardt:
I am actually not happy that we depend on travis for deployment. 
In TeamCity we have different configurations to cover the tests for the application and the deployment. Here in travis deployment problems could cause the build to fail though the app itself works and we show a red label in the GitHub readme.

lauraluiz:
Actually that's the case with the Heroku deployment we have. But I want to pull it out from Travis and build the pipeline directly in Heroku, which I think can also enable deploying PR to review them. So my question would be, why do we create a docker image for each push? It's 2 min that adds up to the travis build and cannot see the usage of it. Can we limit them to master or maybe simply to the tagged versions, which are the ones that are stable enough?

svenmueller:
Ok, i can see a couple of questions which we need to answer
- do we want to use Travis, TC ot both for building/integration/deployment?
- do we want to build a deployable artifact with each build?

Regarding the 1st question:
For me travis seems to be much easier to use, maintain and handle. All resources (scripts, config etc) are in the same repository and not stored somewhere else. If a deployment fails, we should and need to fix it in my opinion.

Regarding the 2nd question:  It would be very helpful to always have an up-to-date Docker image with the latest changes. People want to simply try out changes from the master even if it is not stable yet. Of course, built times should be as short as possible and not slow down the whole process.

svenmueller:
@anas-aso i guess to were looking at an older commit ;)

hajoeichler:
We want Travis to build our open source software. Yes, we want to build a deployable artifact with each build that can be updated on our staging system (heroku) and only on demand we will "release" sunrise with a proper production tag.

jayS-de:
I use automated builds by docker hub. They are triggered after every push to the repository. They also can be automatically tagged. This would remove the need to build the artifacts using travis and improves the build time.

As Hajo mentioned Heroku, this is another construction site, as Heroku has his own image format. Means here we have to create another build pipeline. The good news is, that also Heroku supports automatic building on every push. It also supports release pipelines, with automatic staging deployment and promote from staging to live. It also supports to create automatic deployments for Pull Requests.

svenmueller:
The problem i see when using automated builds by Docker hub is that you can't run tests before an image is created, so e.g. if someone pushes a bad commit, a new Docker image is build nonetheless and maybe then pulled by a Docker client.

jayS-de:
Then use the trigger build webhook provided by docker hub. This can be triggered by travis. Heroku already supports to wait for CI to pass before deployment is started.

jayS-de:
Another option would be to use a release management process like git flow.

lauraluiz:
If we can have it directly in Docker Hub, it would be awesome. I've recently moved Heroku deployment out of Travis to avoid relying on Heroku in the development process (and besides have a cleaner repo and faster build process). It also expects CI to be green in order to deploy the build. We should try to do that for everything. Is it possible @svenmueller to use Docker Hub for that?

schleichardt:
this PR is stale, can we close it?

PR #394, opened 2016-05-18
add request scoped dependency injection and use it for user context o…

schleichardt:
…f checkout shipping

@lauraluiz please review


common/app/common/inject/HttpContextScope.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public <T> Provider<T> scope(final Key<T> key, final Provider<T> provider) {
+        return new Provider<T>() {
+            @Override
+            public T get() {
+                Http.Context currentContext = Http.Context.current();
+                if (currentContext == null) {
+                    Context context = httpContextScopeContext.get();
+                    if (context != null) {
+                        T t = (T) context.map.get(key);
```
lauraluiz: Wouldn't casting here raise an exception when t = NullableObject.INSTANCE?

schleichardt: Interesting!

lauraluiz: I'm confused, because nothing is warning me against it, which makes me think it wouldn't, but how? I don't see it

common/app/common/inject/HttpContextScope.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public String toString() {
+        return "Http.Context.ARGS";
+    }
+
+    private static class Context {
+        final Map<Key, Object> map = Maps.newHashMap();
+        final Lock lock = new ReentrantLock();
+
+        public Context open() {
```
lauraluiz: If this class is no longer implementing an interface, what's the usage of the open/close methods?

common/app/common/inject/HttpContextScope.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        };
+    }
+
+    @Override
+    public String toString() {
+        return "Http.Context.ARGS";
+    }
+
+    private static class Context {
```
schleichardt: hmm, that is a relict to the servlet API. While we are it - maybe the context is never closed?

lauraluiz: Summarizing, what the code does is:
- if there is a current context, it gets the cached element if it exists, or it creates it and stores it if not; in both cases it returns the element.
- if there is no current context, obtains the thread context and then the same: it gets the cached element if it exists, or it creates it and stores it if not; in both cases it returns the element.
So there is clearly repetition which if removed can help a lot to read the code. But I don't dare to touch anything without proper tests. So I'd say, before merging could you create unit tests that contemplate all the different cases and then we proceed to the refactoring of the code?

schleichardt: of course

common/app/common/controllers/SunriseController.java
```diff
@@ -156,21 +156,21 @@ private static Locale currentLocale(final String languageTag, final ProjectConte
         return projectContext.isLocaleAccepted(locale) ? locale : projectContext.defaultLocale();
     }
 
-    private static CountryCode currentCountry(final Http.Session session, final ProjectContext projectContext) {
+    public static CountryCode currentCountry(final Http.Session session, final ProjectContext projectContext) {
```
butenkor: Does it have to be static? If yes and it has to be public, i would put into utility class.

common/app/common/inject/HttpContextScope.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.inject.Scopes;
+import play.mvc.Http;
+
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/*
+
+IMPORTANT: put @ before the route
```
lauraluiz: You mean using non-static controller actions?

common/app/common/controllers/SunriseController.java
schleichardt: Yes, I saw that this could be moved, but one step at a time, now we need to bring the request scoped dependency injection into live then we can move a lot of stuff around.

common/app/common/inject/HttpContextScope.java
lauraluiz: Answer for the future: cast exception only raised when you access a method that the (wrongly) casted object doesn't have.

shopping-cart/app/shoppingcart/common/StepWidgetBean.java
```diff
@@ -1,44 +1,26 @@
 package shoppingcart.common;
 
-public class StepWidgetBean {
+public enum StepWidgetBean {
```
lauraluiz: Nice idea

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressPageController.java
```diff
@@ -24,66 +22,66 @@
...
 
     @RequireCSRFCheck
+    @SuppressWarnings("unused")
     public CompletionStage<Result> process(final String languageTag) {
-        final UserContext userContext = userContext(languageTag);
+        return getOrCreateCart()
+                .thenComposeAsync(this::processAddressForm, HttpExecution.defaultContext());
+    }
+
+    private Result showCheckoutAddressPage(final Cart cart) {
```
lauraluiz: If we extract the logic out of the controller action is fine, even though in some controllers there would be only a `return showThisPage();`, but I think it's an interesting idea to leave the action barely empty and let all the override happen outside these methods.
In any case, I would put these methods as protected so they can be overridden.

lauraluiz: Just to make it clear, but I want to avoid is having different structures per controller. If you know how the structure of a controller looks like, you should be able to know all the other controllers, because all of them will have exactly the same structure. It's a way to be consistent. So for every change, we have to realize all the rest will be looking exactly the same.

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressPageController.java
```diff
@@ -107,71 +105,44 @@ protected boolean formHasErrors(final Form<CheckoutShippingAddressFormData> ship
         return sphere().execute(CartUpdateCommand.of(cart, updateActions));
     }
 
-    protected CompletionStage<Result> handleSuccessfulSetAddress(final UserContext userContext) {
-        final Call call = reverseRouter().showCheckoutShippingForm(userContext.locale().toLanguageTag());
+    protected CompletionStage<Result> handleSuccessfulSetAddress(final Cart cart) {
```
lauraluiz: 👍 if you did it for consistency. I want to keep the same structure for the methods, even though I might not be using those parameters in my implementation. Because the point is to minimize the amount of overridden logic for others and for that I have to provide the typical information you would need.

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressFormBean.java
```diff
@@ -23,6 +21,9 @@
     private AddressFormBean billingAddress;
     private ErrorsBean errors;
 
+    public CheckoutAddressFormBean() {
```
lauraluiz: Oops

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressPageContentFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    protected CheckoutAddressPageContent createWithAddressError(final Form<CheckoutShippingAddressFormData> shippingAddressForm,
+                                                                final Form<CheckoutBillingAddressFormData> billingAddressForm,
+                                                                final ErrorsBean errors) {
+        final CheckoutAddressPageContent pageContent = new CheckoutAddressPageContent();
+        final Address shippingAddress = extractAddress(shippingAddressForm, "Shipping");
+        final Address billingAddress = extractAddress(billingAddressForm, "Billing");
+        final boolean differentBillingAddress = extractBooleanFormField(shippingAddressForm, "billingAddressDifferentToBillingAddress");
+        final CheckoutAddressFormBean formBean = new CheckoutAddressFormBean(shippingAddress, billingAddress, differentBillingAddress, userContext, projectContext, i18nResolver, configuration);
```
lauraluiz: I guess this will be moved to the factory too?

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressFormBean.java
schleichardt: What is wrong?

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressPageContentFactory.java
schleichardt: Yes.

shopping-cart/app/shoppingcart/checkout/address/CheckoutAddressFormBean.java
lauraluiz: That I forgot to create the empty constructor.

lauraluiz:
I have mixed feelings with the factory classes. I really like the idea of having some external class taking care of creating the page for different situations. It can still be easily modified and yet the logic is out of the controller. The negative feeling comes probably from the fact that every time we tried using factory classes, it ended up bad. But hopefully this time is different.

product-catalog/pt/com/commercetools/sunrise/productcatalog/home/SunriseHomeControllerTest.java
```diff
@@ -15,6 +16,7 @@
 
 public class SunriseHomeControllerTest extends WithSunriseApplication {
 
+    @Ignore
     @Test
     public void homeIsAlive() {
         setContext(requestBuilder().build());
```
lauraluiz: @schleichardt can you check the whole commit and fix the tests? It's all about hooks and metrics.
That's the only thing stopping me from merging to master.

common/test/com/commercetools/sunrise/common/ctp/MetricHttpClientTest.java
```diff
@@ -1,65 +1,67 @@
...
 
 public class MetricHttpClientTest {
     private static final HttpRequest SOME_REQUEST = HttpRequest.of(HttpMethod.GET, "somewhere");
     private static final HttpResponse SOME_RESPONSE = HttpResponse.of(200);
 
-    @Test
-    public void executesRequestWithNoContext() throws Exception {
-        executeWithoutContext(SOME_REQUEST, response -> assertThat(response).isEqualTo(SOME_RESPONSE));
-    }
+// TODO Do we still need this?
```
schleichardt: I will remove it, with the new architecture it is always the case that it has the HTTP.Context.

common/test/com/commercetools/sunrise/common/ctp/MetricHttpClientTest.java
```diff
@@ -1,65 +1,67 @@
...
-    private void executeWithContext(final HttpRequest request, Consumer<HttpResponse> test) {
-        Http.Context.current.set(emptyContext());
-        executeWithoutContext(request, test);
-    }
-
-    private void executeWithoutContext(final HttpRequest request, Consumer<HttpResponse> test) {
-        final MetricHttpClient metricHttpClient = MetricHttpClient.of(httpClient());
+    private void testMetrics(final HttpRequest request, BiConsumer<HttpResponse, Http.Context> test) {
+        final Logger metricsLogger = (Logger) LoggerFactory.getLogger(MetricHttpClient.LOGGER_NAME);
+        metricsLogger.setLevel(Level.DEBUG); // TODO check better way to do this
```
schleichardt: I would keep it this way.

product-catalog/pt/com/commercetools/sunrise/productcatalog/home/SunriseHomeControllerTest.java
schleichardt: I just did enough to make the test pass. For making it nice I want to make a pull request so you can review it better.

PR #437, opened 2016-07-13
remove FutureUtils

schleichardt:


lauraluiz:
An easy one, so I review it now ;)
👍

PR #443, opened 2016-07-18
Split sphere client into request scoped and singleton for usage outside controllers

lauraluiz:
Fixes issue on production with failed Injector creation due to missing Http Context

### DIFF
--- a/common/app/com/commercetools/sunrise/common/ctp/CtpModule.java
+++ b/common/app/com/commercetools/sunrise/common/ctp/CtpModule.java
@@ -2,6 +2,7 @@ package com.commercetools.sunrise.common.ctp;
 
 import com.commercetools.sunrise.common.contexts.RequestScoped;
 import com.google.inject.AbstractModule;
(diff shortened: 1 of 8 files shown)


common/app/com/commercetools/sunrise/common/ctp/CtpModule.java
```diff
@@ -20,6 +21,7 @@ protected void configure() {
         bind(HttpClient.class).toInstance(SphereClientFactory.of().createHttpClient());
         bind(SphereClientConfig.class).toProvider(SphereClientConfigProvider.class).in(Singleton.class);
         bind(SphereAccessTokenSupplier.class).toProvider(SphereAccessTokenSupplierProvider.class).in(Singleton.class);
+        bind(SphereClient.class).annotatedWith(Names.named("singleton")).toProvider(SphereClientProvider.class).in(Singleton.class);
```
lauraluiz: @schleichardt after 5 seconds this is the first name that came to my mind, suggestions?

PR #445, opened 2016-07-20
Documentation 0.8.0

schleichardt:
@lauraluiz please review, In this pull request I want to provide the documentation for the new stuff in 0.8.0 or earlier, moving it to the right place would be your task


staged-documentation.md
```diff
@@ -0,0 +1,282 @@
...
+    //do not call here super.create because you would get a bean of the original class and not the subclass
+    @Override
+    public ProductDetailPageContent create(final ProductProjection product, final ProductVariant variant) {
+        final MyProductDetailPageContent bean = new MyProductDetailPageContent();
+        initialize(bean, product, variant);
+        return bean;
+    }
+
+    //the new initialize method contains as first parameter the bean of your subclass and then the parameters of the create method
+    protected final void initialize(final MyProductDetailPageContent bean, final ProductProjection product, final ProductVariant variant) {
```
lauraluiz: I know this is the proper way of doing it if you want to enable subclasses. But it's easier to explain/understand if you just put the missing information in the `create` method. If they plan on using subclasses they will probably realize how to fix it on their own or they will check our class as inspiration (since they know it is possible to do because we are doing it). I would try to keep the explanations as simple as possible (as in the trainings), it gives the feeling it is very simple and leaves a positive impression. We can always comment in text later that a good practice to allow subclasses would be to follow the same approach as we do, but that it is unnecessary for simple cases.

lauraluiz: I mean it especially because of the "do not override initialize" and "call super.inititalize". That can be confusing and get in the way of the topic explained.

staged-documentation.md
```diff
@@ -0,0 +1,282 @@
...
+
+```
+
+To make Play know the module, add it to `application.conf`:
+
+```
+play.modules.enabled += "absolute.path.to.your.FactoryModule"
+```
+
+The final home page then looks like this:
```
lauraluiz: Is there supposed to be a picture here? :P

staged-documentation.md
schleichardt: I also tried to be simplistic by showing just the one way to fit them all instead of two. So should you or me add the simpler way?

schleichardt: you can also remove confusing comments

staged-documentation.md
schleichardt: I copied a template and may forgot to remove this, do we need a pic here?

staged-documentation.md
lauraluiz: No, I think the comments help a lot. I would simply avoid creating this `initialize` method, so that they don't have to be careful not to override the `super.initialize`or call the same method infinitely. If you agree, I can change it myself.

staged-documentation.md
lauraluiz: Not to me, specially since we don't show how it looks before. Also, I think it's not an example where the visual part is important.

staged-documentation.md
```diff
@@ -0,0 +1,282 @@
...
+
+    @Override
+    public void acceptPageData(final PageData pageData) {
+        final ComponentBean component = createComponentBean();
+        pageData.getContent().addComponent(component);
+    }
+
+    private ComponentBean createComponentBean() {
+        final ComponentBean component = new ComponentBean();
+        //the template is in conf/templates/components/summercampaign/banner.hbs or of course your theme
```
lauraluiz: I think there's a typo here (or I don't understand the sentence)

staged-documentation.md
```diff
@@ -0,0 +1,282 @@
...
+    @Provides
+    public MultiControllerComponentResolver multiControllerComponentResolver() {
+        return new MultiControllerComponentResolverBuilder()
+                .add(SummerCampaignControllerComponent.class, controller -> !controller.getFrameworkTags().contains("checkout"))
+                .build();
+    }
+}
+```
+
+The `ComponentsModule` can be reused for multiple components, just add a `add` line for building the `MultiControllerComponentResolver`.
```
lauraluiz: Now it sounds a bit confusing, but I'm not sure what you want to say. I will just say what I understand:

1) The `ComponentsModule` is a good place to put all bindings related to components. But actually I don't think we would have any normal binding in `configure`, right? We will only have the `MultiControllerComponentResolver` binding (at least in the current state).
2) It would be good to explain somewhere the 2 different ways to register components (per controller vs. in the `MultiControllerComponentResolver`), it was a question that popped up during the training.
3) To avoid repeating the "add", I would use "insert", "put" or separate more both "add"s.

Overall if my assumptions are right, I would propose:
"With this Guice Module we can register any component to multiple controllers with a single line of code. You just need to add your component to the `MultiControllerComponentResolver`, which can be easily achieved with its builder."
I would even put this text before the example.

staged-documentation.md
```diff
@@ -0,0 +1,282 @@
...
+
+Don't forget to register the module in `application.conf` with
+
+```
+play.modules.enabled += "absolute.path.to.your.ComponentsModule"
```
lauraluiz: I would create have already this Module created in the Demo project or "Blank". Then the question is whether we want this documentation to assume the module is created or not.

staged-documentation.md
schleichardt: I agree.

staged-documentation.md
schleichardt: So I delete this sentence in a commit.

staged-documentation.md
schleichardt: Can you apply the fix?

staged-documentation.md
schleichardt: Yes, that can be an ongoing discussion. Maybe it also makes sense to merge the modules for factories and components, because components may take part in using the overridden factories.

staged-documentation.md
```diff
@@ -0,0 +1,282 @@
...
+<hr class="home-suggestions-hr" />
+```
+
+And then it looks like
+![result](documentation-images/summercampaign-sungrasses-home.png)
+
+
+## Handlebars adding helpers
+
+*We suggest to not add handlebars helpers to use for calculations or logic, this is not what is handlebars designed for and selected for.*
```
lauraluiz: "We strongly suggest not to add Handlebars helpers to enable calculations or any kind of business logic in the templates. Handlebars helpers are only meant to simplify presentation logic (in contrast of a pure logic-less template such as [Mustache](https://mustache.github.io/)). Using helpers for other purposes would defeat the logic-less approach that we try to promote."

lauraluiz:
@schleichardt you can do a last review. I split the documentation on topics per page and put them inside the `manual` folder (which has a folder for images). On the way I also fixed some things I saw, I realised we have to put the Guidelines and Contributing pages up to date :D

GUIDELINES.md
```diff
@@ -4,121 +4,121 @@ General guidelines about how to successfully develop an application based on [co
...
 #####Take care of
-- Only aware of other models.
+* Only aware of other models.
 
 #####Patterns and rules
-- Immutable objects.
-- Highly reused.
-- Created by static factory methods or builders.
-- Should not end with the term "Model".
+* Immutable objects.
```
schleichardt: hmm, we use now beans

GUIDELINES.md
```diff
@@ -4,121 +4,121 @@ General guidelines about how to successfully develop an application based on [co
...
+* Highly reused.
+* Created by static factory methods or builders.
```
schleichardt: with beans instead of immutable objects, the beans can be created directly

GUIDELINES.md
lauraluiz: Yes, that's why we have to put them up-to-date. I just fixed formatting and barely nothing else. I'd say, let's just ignore it and fix it in this issue: https://github.com/commercetools/commercetools-sunrise-java/issues/450

CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@ If you have push access to the repository you can fix them directly otherwise ju
 1. fork the repository
 1. produce production code and unit tests
 1. make a pull request
-1. metion and/or assign @lauraluiz and @schleichardt, who will review and optionally merge
+1. mention and/or assign @lauraluiz and @schleichardt, who will review and optionally merge
```
lauraluiz: 😅

lauraluiz:
You are still reviewing, right @schleichardt ? Just to make sure, maybe we both are waiting for nothing :D

README.md
```diff
@@ -7,7 +7,7 @@ Sunrise Java :sunrise:
 The next generation shop template.
 
 * Demo: https://sunrise.commercetools.com
-* [Documentation](https://github.com/commercetools/commercetools-sunrise-java/blob/master/ONBOARDING.md)
+* [Documentation](https://github.com/commercetools/commercetools-sunrise-java/blob/master/manual/)
```
schleichardt: I would prefer not absolute links for this, so we can use the links in branches
```
[Documentation](manual/)
```

also works

schleichardt:
👍

README.md
lauraluiz: Nice to know! I didn't give it much thought, I just modified it. I'll change it.

PR #463, opened 2016-07-26
Added global filter for CSRF check

ccsalazarr-commercetools:
Added global filter and removed annotations per action.

### DIFF
--- a/common/app/com/commercetools/sunrise/common/basicauth/BasicAuthHttpFilters.java
+++ b/common/app/com/commercetools/sunrise/common/basicauth/BasicAuthHttpFilters.java
@@ -1,5 +1,6 @@
 package com.commercetools.sunrise.common.basicauth;
 
+import play.filters.csrf.CSRFFilter;
 import play.http.HttpFilters;
 import play.mvc.EssentialFilter;
 
@@ -9,13 +10,13 @@ import javax.inject.Inject;
  * Based on play.mvc.http.DefaultHttpFilters with fix https://github.com/playframework/playframework/pull/6238, which will be released with Play 2.5.5.
  * REMOVE AS SOON AS IT IS RELEASED!
  */
-public class BasicAuthHttpFilters implements HttpFilters {
+public class BasicAuthHttpFilters implements HttpFilters{
 
     private final EssentialFilter[] filters;
 
     @Inject
-    public BasicAuthHttpFilters(final BasicAuthFilter basicAuthFilter) {
-        this.filters = new EssentialFilter[] {basicAuthFilter.asJava()};
+    public BasicAuthHttpFilters(final BasicAuthFilter basicAuthFilter, final CSRFFilter csrfFilter) {
+        this.filters = new EssentialFilter[] {basicAuthFilter.asJava(), csrfFilter.asJava()};
     }
 
     @Override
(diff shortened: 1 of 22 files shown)


lauraluiz:
@ccsalazarr-commercetools looks good! Have you checked all forms still work properly? Unfortunately we don't have webtests yet to check that for us :(

common/app/com/commercetools/sunrise/common/basicauth/BasicAuthHttpFilters.java
```diff
@@ -9,13 +10,13 @@
  * Based on play.mvc.http.DefaultHttpFilters with fix https://github.com/playframework/playframework/pull/6238, which will be released with Play 2.5.5.
  * REMOVE AS SOON AS IT IS RELEASED!
  */
-public class BasicAuthHttpFilters implements HttpFilters {
+public class BasicAuthHttpFilters implements HttpFilters{
```
lauraluiz: Not the best name now, but since this is a fix and will be fixed soon, I don't want to provide double breaking change.
What I'm not sure now if you can have multiple HttpFilters in your configuration. Have you tried? If not, try to create an additional filter `CsrfHttpFilters` using this one as base. `BasicAuthHttpFilters` will remain as it was before, and `CsrfHttpFilters` will look the same but with the `CSRFFilter` instead. Then you add this one to `application.conf` (be careful with using `+=` instead of `=`.

common/app/com/commercetools/sunrise/play/SunriseHttpFilters.java
```diff
@@ -9,13 +11,13 @@
  * Based on play.mvc.http.DefaultHttpFilters with fix https://github.com/playframework/playframework/pull/6238, which will be released with Play 2.5.5.
  * REMOVE AS SOON AS IT IS RELEASED!
  */
-public class BasicAuthHttpFilters implements HttpFilters {
+public class SunriseHttpFilters implements HttpFilters {
```
lauraluiz: @ccsalazarr-commercetools I've finally renamed it to `SunriseHttpFilters`, as I saw it is unique per project (or at least that's how it is documented). I've also move it outside from `basicauth`, since it's not related to it.

common/app/com/commercetools/sunrise/play/SunriseHttpFilters.java
```diff
@@ -1,5 +1,7 @@
-package com.commercetools.sunrise.common.basicauth;
+package com.commercetools.sunrise.play;
```
lauraluiz: @schleichardt I've created the `play` package, where I pretend to put very Play-related things. For example, in this case it's the `Filters` or the `ApplicationLoader` that are used to alter the behaviour of Play. Open to suggestions.

schleichardt: :D so we use different strategies, I used packages like play would have like, so I would have put it into a package filters.

play is too general and will contain finally a lot of classes with very different types and purposes

lauraluiz: You are wrong, I also use the strategy you mention. But in this case it's different.

What I want to achieve with this is not having so many packages in `common` with just one class inside, especially since all these classes are unique and won't have more `Filters`, more `Configuration` or more `ApplicationLoader`. They are just replacing the default stuff in Play. If you tell me there are at least two filters/configuration/application-loaders related classes, then I'll agree with you.

Of course then we have `BasicAuth` which is using the filter strategy and could go inside the `filters` package, but this is just the current implementation, just a few weeks ago it was using another strategy. That's why I would like to keep this outside any folder referencing `filters`.

schleichardt: agreed :+1:

PR #464, opened 2016-07-26
Page meta via components & reflection reverse routes, closes #431, closes #421

### LINKED ISSUE #421: PageMetaFactoryImpl needs to be modularized sucks all routers into

### LINKED ISSUE #431: Reverse router less invasive
- Move all methods to its own interface (grouped).
- Use the particular interface in controllers and everywhere.
- Where to declare the routes for the theme (PageMetaFactory vs controllers vs other place).
- Improve TestableReverseRouter (is it necessary?).

schleichardt:


schleichardt:
@lauraluiz please review, but do not merge, I just want your feedback if this is good enough to continue.

schleichardt:
@lauraluiz it is all yours

schleichardt:
these routes have not migrated yet: HomeReverseRouter, ProductReverseRouter, CheckoutReverseRouter, AddressBookReverseRouter, SetupReverseRouter

schleichardt:
I added a thing to use "GET /" by default, so for example devs choose not to use every checkout step, like skipping shipping method, but gives a default it does not explode

AndrewHK:
Great that you guys are changing this. That would affect all of our project though. Right ?

lauraluiz:
@AndrewHK yep, it will have an impact on your code. We will try to provide info about how to migrate so that it's less painful :)

PR #466, opened 2016-07-26
Replace CMS classes with CMS module

lauraluiz:
@schleichardt please review


build.sbt
```diff
@@ -82,6 +82,17 @@ lazy val commonSettings = Release.publishSettings ++ Seq (
   )
 )
 
+lazy val sunriseModuleDependencies = Seq (
```
schleichardt: this is duplicated, we could just have directly in build.sbt and it would be applied to all modules:

```scala
resolvers in ThisBuild ++= Seq (
     Resolver.sonatypeRepo("releases"),
     Resolver.sonatypeRepo("snapshots"),
     Resolver.mavenLocal
   )
```

schleichardt:
:+1:

build.sbt
lauraluiz: You're right, in this case it makes no sense to put it for a particular set of dependencies anymore. I've pulled it up.

PR #480, opened 2016-08-03
Create cart when needed, closes #420

### LINKED ISSUE #420: lazy cart
do not create a cart until needed

ccsalazarr-commercetools:


lauraluiz:
👍

common/app/com/commercetools/sunrise/shoppingcart/CartBeanFactory.java
```diff
@@ -16,12 +17,27 @@ public CartBean create(final Cart cart) {
         return bean;
     }
 
+    public CartBean createWithEmptyCart() {
```
schleichardt: why the `With` in the name?

schleichardt: in CartDetailPageContent the parameter Cart is nullable, so why there are two different styles?

PR #488, opened 2016-08-10
Handle session framework, closes #309

### LINKED ISSUE #309: Find a better way to handle sessions
E.g. some sessions are dependent from each other (customer -> cart). And they usually belong to different modules.
- Think frameworky.
- How to deal with dependent info (logout removes cart & customer).

lauraluiz:


common/app/com/commercetools/sunrise/common/controllers/SunriseFrameworkController.java
```diff
@@ -259,4 +263,12 @@ protected final void saveUnexpectedFormError(final Form<?> form, final Throwable
         form.reject("Something went wrong, please try again"); // TODO i18n
         logger.error("The CTP request raised an unexpected exception", throwable);
     }
+
+    protected final void overwriteCartInSession(final Cart cart) {
```
schleichardt: I am not happy with putting it in the core, I would prefer that this is also solved via components which are called when a primary cart is loaded/updated/created

common/app/com/commercetools/sunrise/common/controllers/SunriseFrameworkController.java
```diff
@@ -259,4 +263,12 @@ protected final void saveUnexpectedFormError(final Form<?> form, final Throwable
...
+        injector().getInstance(CartSessionHandler.class).overwriteInSession(session(), cart);
+    }
+
+    protected final void overwriteCustomerInSession(final Customer customer) {
```
schleichardt: same here

common/app/com/commercetools/sunrise/common/pages/PageMetaFactoryImpl.java
```diff
@@ -83,4 +87,9 @@ private PageMeta getPageMeta(final UserContext userContext, final Http.Context c
         return pageMeta;
     }
 
+    private void fillUser(final PageMeta pageMeta, final Http.Session session) {
+        injector.getInstance(CustomerSessionHandler.class).findUserInfo(session)
```
schleichardt: for this thing I would use a component, maybe a herold component

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
```diff
@@ -0,0 +1,69 @@
...
+import io.sphere.sdk.json.SphereJsonUtils;
+import org.slf4j.Logger;
+import play.libs.Json;
+import play.mvc.Http;
+
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.Optional;
+
+public abstract class SessionHandlerBase<T> implements SessionHandler<T> {
```
schleichardt: Is handler the right concept? I am not sure if it is some kind of registry. http://martinfowler.com/eaaCatalog/registry.html

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public void removeFromSession(final Http.Session session) {
+        removeRelatedValuesFromSession(session);
+    }
+
+    protected abstract void overwriteRelatedValuesInSession(final Http.Session session, final T value);
+
+    protected abstract void removeRelatedValuesFromSession(final Http.Session session);
+
+    protected Optional<String> findValue(final Http.Session session, final String key) {
```
schleichardt: maybe `findValueByKey`? The name `findValue` sounds to me like you put the thing you search for as argument not the key.

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
```diff
@@ -0,0 +1,69 @@
...
+        return Optional.ofNullable(session.get(key));
+    }
+
+    protected <U> Optional<U> findValue(final Http.Session session, final String key, final Class<U> clazz) {
+        return findValue(session, key)
+                .flatMap(valueAsJson -> {
+                    try {
+                        return Optional.of(SphereJsonUtils.readObject(valueAsJson, clazz));
```
schleichardt: This is not s cool thing, the SphereJsonUtils is really just for commercetools objects, do not use it als JSON library for other stuff.

We have also the class com.commercetools.sunrise.common.utils.JsonUtils which could be used as helper thing plus there is a Play util: https://www.playframework.com/documentation/2.5.x/api/java/play/libs/Json.html

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
```diff
@@ -0,0 +1,69 @@
...
+    protected void removeValues(final Http.Session session, final List<String> keys) {
+        keys.forEach(key -> removeValues(session, key));
+    }
+
+    protected void removeValues(final Http.Session session, final String key) {
+        session.remove(key);
+        logger().debug("Removed from session \"{}\"", key);
+    }
+
+    protected abstract Logger logger();
```
schleichardt: Does it make sense to default implement it like:

```java
protected Logger logger() {
  return LoggerFactory.getLogger(this.getClass());
}
```

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
```diff
@@ -0,0 +1,69 @@
...
+                    try {
+                        return Optional.of(SphereJsonUtils.readObject(valueAsJson, clazz));
+                    } catch (JsonException e) {
+                        logger().error("Could not parse value in session key \"{}\"", key, e);
+                        return Optional.empty();
+                    }
+                });
+    }
+
+    protected void overwriteValue(final Http.Session session, final String key, final String value) {
```
schleichardt: can you rename or inline it?

It can bring some surprises since `protected <U> void overwriteValue(final Http.Session session, final String key, final U value)` has quite a similar signature.

common/app/com/commercetools/sunrise/myaccount/CustomerSessionHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+import javax.inject.Inject;
+import java.util.Optional;
+
+import static java.util.Arrays.asList;
+
+public class CustomerSessionHandler extends SessionHandlerBase<Customer> {
+
+    private static final Logger logger = LoggerFactory.getLogger(CustomerSessionHandler.class);
+    private static final String CUSTOMER_ID_SESSION_KEY = "sunrise-customer-id";
```
schleichardt: does it make sense to get these from application.conf?

schleichardt: of course we could provide defaults here or in reference.conf

common/app/com/commercetools/sunrise/shoppingcart/CartFinderBySession.java
```diff
@@ -15,25 +14,25 @@
 import java.util.concurrent.CompletionStage;
 import java.util.function.UnaryOperator;
 
-import static com.commercetools.sunrise.shoppingcart.CartSessionUtils.overwriteCartSessionData;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 
 public class CartFinderBySession implements CartFinder<Http.Session> {
```
schleichardt: this class is unused

schleichardt:
Actually I would solve this differently.
I would register a request scoped component (much like CustomerSessionHandler) which is saving the customer data in the session on the event of log in/update of the customer. And with DI a component or a controller can get this component to read the customer.

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
lauraluiz: You're totally right, I'm changing it.

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
lauraluiz: Honestly I don't know why I used it. I'm using the Play helper now.

common/app/com/commercetools/sunrise/common/sessions/SessionHandlerBase.java
lauraluiz: Wouldn't it be using the logger of SessionHandlerBase then? Because that wouldn't be a good option and it would be forgotten to override the method.

common/app/com/commercetools/sunrise/myaccount/CustomerSessionHandler.java
lauraluiz: Yes, it could be useful for a quite advanced feature, but I don't really see the need now. Session key conflict could be the only one but I see that very unlikely. Which situation do you have in mind?

common/app/com/commercetools/sunrise/shoppingcart/CartFinderBySession.java
lauraluiz: Yes, it is now. I think I didn't dare to replace the code that takes care of that with this because there were ongoing changes in that class at that moment. Now it's safe to be replaced.

common/app/com/commercetools/sunrise/myaccount/CustomerSessionHandler.java
lauraluiz: But currently you could simply do it programatically (extend the class and overwrite the session key method) and bind your session class instead, so it's not impossible to do. It's just simply not supported via configuration yet. So instead of programatically I can switch it to configuration, but it was possible all the time.

common/app/com/commercetools/sunrise/common/sessions/SessionHandler.java
```diff
@@ -0,0 +1,12 @@
...
+
+import play.mvc.Http;
+
+import javax.annotation.Nullable;
+
+public interface SessionHandler<T> {
+
+    void overwriteInSession(final Http.Session session, @Nullable final T value);
+
+    void removeFromSession(final Http.Session session);
```
lauraluiz: Put read method

PR #499, opened 2016-08-19
Remove Request Context Module

lauraluiz:
@schleichardt


common/app/com/commercetools/sunrise/common/injection/SunriseModule.java
```diff
@@ -1,19 +1,50 @@
...
+        bind(RequestContext.class).toProvider(RequestContextProvider.class).in(RequestScoped.class);
+        bind(RequestHookContext.class).to(RequestHookContextImpl.class).in(RequestScoped.class);
         bind(MultiControllerComponentResolver.class)
                 .annotatedWith(Names.named("controllers"))
                 .toProvider(RoutesMultiControllerComponentResolverProvider.class)
                 .in(Singleton.class);
     }
+
+    @Provides
+    @RequestScoped
```
lauraluiz: Cons: Duplicated code with the test
Pros: One class less and it kind of calls this only once, yet keeps the request scope.
Do you have any clue why this is never called again? But it behaves differently than when I use `@Singleton`, besides everything works perfectly.

app/Module.java
```diff
@@ -16,4 +24,16 @@ protected void configure() {
         bind(TemplateEngine.class).toProvider(HandlebarsTemplateEngineProvider.class).in(Singleton.class);
         bind(I18nResolver.class).toProvider(ConfigurableI18nResolverProvider.class).in(Singleton.class);
     }
+
+    @Provides
+    public MultiControllerComponentResolver foo() {
```
schleichardt: your opportunity to fix the name foo

common/app/com/commercetools/sunrise/common/injection/SunriseModule.java
schleichardt: I think it is a beautiful simplistic solution. Good job.

conf/application.conf
```diff
@@ -317,9 +317,6 @@ play.crypto.secret=${?APPLICATION_SECRET}
 play.application.loader = "com.commercetools.sunrise.play.inject.SunriseGuiceApplicationLoader"
 play.http.filters = "com.commercetools.sunrise.play.http.SunriseHttpFilters"
 
-play.modules.enabled += "com.commercetools.sunrise.common.contexts.RequestScopeModule"
```
schleichardt: 👍

schleichardt:
👍

PR #503, opened 2016-08-23
Rename primary cart

lauraluiz:
@schleichardt please review, I have simplified as well the names for the `require` methods.


schleichardt:
+1

PR #521, opened 2016-10-10
Remove DI Modules

lauraluiz:
@schleichardt finally, all Modules removed. This time I didn't think much about deep improvements, just a few things to make it simpler. Please review.


app/Module.java
```diff
@@ -10,22 +14,58 @@
...
 import javax.inject.Singleton;
+import java.util.Locale;
+
+import static java.util.Collections.singletonList;
 
 public class Module extends AbstractModule {
 
     @Override
     protected void configure() {
+        bind(SphereClient.class).toProvider(RequestScopedSphereClientProvider.class).in(RequestScoped.class);
```
schleichardt: why is this not in the other module?

schleichardt:
looks good

app/Module.java
lauraluiz: Basically the rule of thumb to decide whether to put a binding in this module or in SunriseModule is how likely is that the developer would like to override this binding. In SunriseModule is not possible to override without having to create your own SunriseModule with the desired changes, whereas Module is supposed to be in the project itself so you just have to modify the code.
Therefore I've put in SunriseModule the `global` SphereClient because it's a special client that people shouldn't need to override. But this one, is very likely that they'll want to add some metrics or wrap it for some other purpose.

PR #538, opened 2016-11-10
Rework session classes

lauraluiz:
@schleichardt here it is


common/app/com/commercetools/sunrise/common/injection/SunriseModule.java
```diff
@@ -49,6 +49,12 @@ protected void configure() {
         return Http.Context.current();
     }
 
+    @Provides
+    @RequestScoped
+    public Http.Session httpSession(final Http.Context httpContext) {
+        return httpContext.session();
+    }
```
lauraluiz: Makes testing easier, since classes that need session don't need to inject the whole `Http.Context`  for each test.

common/app/com/commercetools/sunrise/common/sessions/TypedSessionStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.google.inject.ImplementedBy;
+
+import java.util.Optional;
+
+/**
+ * Represents a strategy used to read and write objects into the user's session.
+ */
+@ImplementedBy(SerializableSessionCookieStrategy.class)
+public interface TypedSessionStrategy extends SessionStrategy {
```
lauraluiz: I didn't like `ObjectSessionStrategy` but couldn't find a replacement. I'm testing my luck with `TypedSessionStrategy`, as defining which type are you storing is important here. But I don't really like it either, it sounds as String is not a type. Help wanted here :(

lauraluiz:
I've put some comments over stuff that might need some other explanations. Anyway, check specially all naming of classes and methods from the session classes.
There's a lot of noise from places where session is used, but you can easily skip those files.

common/app/com/commercetools/sunrise/common/injection/SunriseModule.java
```diff
@@ -49,6 +49,12 @@ protected void configure() {
         return Http.Context.current();
     }
 
+    @Provides
+    @RequestScoped
+    public Http.Session httpSession() {
+        return httpContext().session();
+    }
```
lauraluiz: Makes testing easier, since classes that need session don't need to inject the whole Http.Context for each test.

common/app/com/commercetools/sunrise/common/sessions/CacheableSessionCookieStrategy.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static final String DEFAULT_SESSION_ID_KEY = "sunrise-session-id";
+    private final String sessionIdKey;
+    private final CacheApi cacheApi;
+
+    @Inject
+    public CacheableSessionCookieStrategy(final Http.Session session, final CacheApi cacheApi, final Configuration configuration) {
+        super(session);
+        this.cacheApi = cacheApi;
+        this.sessionIdKey = configuration.getString("session.id", DEFAULT_SESSION_ID_KEY);
```
schleichardt: I am surprised that this is not called something like `session.idKey`

common/app/com/commercetools/sunrise/common/sessions/StorableResource.java
```diff
@@ -0,0 +1,24 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Enables a particular object to be stored somewhere.
+ * The interface does not define where (e.g. session, cache) or how (e.g. the entire object, only specific parts)
+ * the object is stored, which should be specified by the implementation class.
+ * @param <T> Class of the stored object
+ */
+public interface StorableResource<T> {
```
schleichardt: If it is called `StorableResource` then sounds like the object that will be stored needs to extend this interface. `ResourceStore` would fit much better.

common/app/com/commercetools/sunrise/common/sessions/TypedSessionStrategy.java
schleichardt: `ObjectStoringSessionStrategy`?

common/app/com/commercetools/sunrise/myaccount/CustomerFinderBySession.java
```diff
@@ -6,33 +6,32 @@
...
     @Inject
-    private CustomerSessionHandler customerSessionHandler;
+    private CustomerInSession customerInSession;
     @Inject
     private RequestHookContext hookContext;
 
     @Override
-    public CompletionStage<Optional<Customer>> findCustomer(final Http.Session session) {
-        final CompletionStage<Optional<Customer>> customerStage = fetchCustomer(session);
+    public CompletionStage<Optional<Customer>> findCustomer(final Void unused) {
```
schleichardt: The interface CustomerFinder does not extend anything so you can remove the parameter.

```
public interface CustomerFinder<C> {

    CompletionStage<Optional<Customer>> findCustomer(final C customerIdentifier);
}
```

common/app/com/commercetools/sunrise/common/sessions/CacheableSessionCookieStrategy.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Uses a session cookie to store string information about the user.
+ * Objects are first saved into the cache and uniquely associated with the user's session cookie with a generated session ID.
+ * For more information: <a href="https://www.playframework.com/documentation/2.5.x/JavaSessionFlash">Session in Play Framework</a>
+ */
+@RequestScoped
+public class CacheableSessionCookieStrategy extends SessionCookieStrategy implements TypedSessionStrategy {
```
schleichardt: Maybe rename it to `ObjectInCacheSessionCookieStategy`?

common/app/com/commercetools/sunrise/common/sessions/SerializableSessionCookieStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.inject.Inject;
+import java.util.Optional;
+
+/**
+ * Uses a session cookie to store string information about the user.
+ * Objects are first serialized to JSON to be able to write them into the session cookie.
+ * Notice that cookies have a very limited space, so do not use this strategy to store a large amount of data.
+ */
+@RequestScoped
+public class SerializableSessionCookieStrategy extends SessionCookieStrategy implements TypedSessionStrategy {
```
schleichardt: Mayve `ObjectInSessionSessionCookieStrategy`?

common/app/com/commercetools/sunrise/myaccount/CustomerFinderBySession.java
lauraluiz: This particular implementation of CustomerFinder does not need anything from the developer because it just uses the injected CustomerInSession instance. But that won't always be the case. You could expect the ID, email, external key, orderID... there can be multiple implementations. That's why the parameter is kept there and in the only subclass so far.

common/app/com/commercetools/sunrise/common/sessions/CacheableSessionCookieStrategy.java
lauraluiz: Good catch, I gave it a quick name to proof the concept and then I forgot about it!

PR #540, opened 2016-11-11
Fix product with no categories suggestion issue

lauraluiz:
@mmoelli @jayS-de I had a test to cover the other method (for categories) but for some reason missed the one for products... even though I remember clearly it was the case I wanted to cover 😅 

Anyway, here a quick and effective fix that I will deploy as soon as Travis decides to work again.


lauraluiz:
@schleichardt please review

common/test/com/commercetools/sunrise/common/suggestion/SunriseProductRecommendationTest.java
```diff
@@ -34,4 +55,128 @@ protected void configure() {
         });
         return injector.getInstance(SunriseProductRecommendation.class);
     }
+
+    private ProductProjection productWithNoCategories() {
```
schleichardt: A lot of code to express empty categories, I would prefer:

```java
final ProductProjection productProjection = 
        SphereJsonUtils.readObject("{\"categories\": [], \"variants\": []}", ProductProjection.class);
```

PR #547, opened 2016-11-23
Set customer email to cart, closes #546

### LINKED ISSUE #546: SunriseAddProductToCartController can't find existing cart, CartDetailController shows wrong cart
With framework v.0.14.0, we experience major errors (see also #545). This effects also the cart handling and it seems to be produced by the following bugs: 

1. when a user is signed in and adds a product to a cart, the SunriseAddProductToCartController searches for a cart id in the session. Since this id is never stored in the session, it creates always a new cart for the customer. As a workaround, we override the handleSuccessfulAction() method and add the created cart ourself by calling `cartInSession.store(cart);`

2. This however doesn't fix the cart view (SunriseCartDetailController). Here, the CartFinderBySession is asked for a cart by customer id but never returns the cart belonging to the current customer. I'm not sure where the cause is though, maybe the newly created cart doesn't get a customerId. We also did a workaround for now by implementing our own CartFinderBySession with the following method buildQuery():

`private Optional<CartQuery> buildQuery() {`
`        return cartInSession.findCartId()`
`            .map(this::buildQueryByCartId)`
`            .map(query -> query`
`                .plusPredicates(cart -> cart.cartState().is(CartState.ACTIVE))`
`                .plusExpansionPaths(c -> c.shippingInfo().shippingMethod()) // TODO use run hook on cart query instead`
`                .plusExpansionPaths(c -> c.paymentInfo().payments())`
 `               .withSort(cart -> cart.lastModifiedAt().sort().desc())`
 `               .withLimit(1));`
 `   }`

It would be nice to fix these bugs as we are expecting this to be standard behaviour.

lauraluiz:
@schleichardt please could you review it?

Closes #546

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -175,6 +175,7 @@ def configCommonTestSettings(scopes: String) = Seq(
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-v"),
   libraryDependencies ++= Seq (
     "org.assertj" % "assertj-core" % "3.0.0" % scopes,
+    "com.commercetools.sdk.jvm.core" % "commercetools-test-lib" % "1.0.0-RC2" % scopes,
     PlayImport.component("play-test") % scopes
   ),
   dependencyOverrides ++= Set (
--- a/common/app/com/commercetools/sunrise/shoppingcart/CartComponent.java
+++ b/common/app/com/commercetools/sunrise/shoppingcart/CartComponent.java
@@ -15,16 +15,23 @@ import com.neovisionaries.i18n.CountryCode;
 import io.sphere.sdk.carts.Cart;
 import io.sphere.sdk.carts.CartDraft;
 import io.sphere.sdk.carts.CartDraftBuilder;
+import io.sphere.sdk.carts.CartDraftDsl;
 import io.sphere.sdk.carts.commands.CartCreateCommand;
 import io.sphere.sdk.carts.commands.CartUpdateCommand;
 import io.sphere.sdk.carts.commands.updateactions.SetCountry;
+import io.sphere.sdk.carts.commands.updateactions.SetCustomerEmail;
 import io.sphere.sdk.carts.commands.updateactions.SetShippingAddress;
 import io.sphere.sdk.client.SphereClient;
+import io.sphere.sdk.commands.UpdateAction;
+import io.sphere.sdk.customers.Customer;
 import io.sphere.sdk.customers.CustomerSignInResult;
 import io.sphere.sdk.expansion.ExpansionPathContainer;
 import io.sphere.sdk.models.Address;
+import play.libs.concurrent.HttpExecution;
 
 import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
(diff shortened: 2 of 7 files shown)


common/it/com/commercetools/sunrise/common/TestFixtures.java
```diff
@@ -0,0 +1,77 @@
...
+            client.executeBlocking(deleteFunction.apply(resource));
+        } catch (ConcurrentModificationException e) {
+            final T resourceWithCorrectVersion = client.executeBlocking(fetchSupplier.get());
+            if (resourceWithCorrectVersion != null) {
+                client.executeBlocking(deleteFunction.apply(resourceWithCorrectVersion));
+            }
+        } catch (NotFoundException e) {
+            // mission indirectly accomplished
+        }
+    }
```
lauraluiz: @schleichardt not sure if this is the best strategy. I want to avoid leaving zombie resources, so I tried doing this. Another alternative I thought of is the "force delete flag" that was mentioned sometimes, but I think it was never implemented, right?
How do you deal with it? Couldn't find any significant code in the JVM SDK.

build.sbt
```diff
@@ -175,6 +175,7 @@ def configCommonTestSettings(scopes: String) = Seq(
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-v"),
   libraryDependencies ++= Seq (
     "org.assertj" % "assertj-core" % "3.0.0" % scopes,
+    "com.commercetools.sdk.jvm.core" % "commercetools-test-lib" % "1.0.0-RC2" % scopes,
```
schleichardt: Hmm, I'd rather suggest we create a test lib that can be used everywhere, this one is still tailored for the JVM SDK.

common/it/com/commercetools/sunrise/common/TestFixtures.java
schleichardt: see io.sphere.sdk.products.ProductFixtures#deleteWithRetry

```java
private static void deleteWithRetry(final BlockingSphereClient client, final Versioned<Product> unPublishedProduct, final int ttl) {
    if (ttl > 0) {
        try {
            client.executeBlocking(ProductDeleteCommand.of(unPublishedProduct));
        } catch(final ConcurrentModificationException e) {
            final Versioned<Product> versioned = Versioned.of(unPublishedProduct.getId(), e.getCurrentVersion());
            deleteWithRetry(client, versioned, ttl - 1);
        }
    } else {
        throw new RuntimeException("cannot delete product due to too much concurrent updates, product: " + unPublishedProduct);
    }
}
```

schleichardt: No, force delete is still a dream.

common/it/com/commercetools/sunrise/common/WithSphereClient.java
```diff
@@ -44,7 +44,7 @@ protected static String clientSecret() {
         return getValueForEnvVar(IT_CTP_CLIENT_SECRET);
     }
 
-    private synchronized static BlockingSphereClient sphereClient() {
+    protected synchronized static BlockingSphereClient sphereClient() {
```
schleichardt: No, I would not reuse this from the JVM SDK because there is too much surprises hidden.

build.sbt
schleichardt: I would not use this early release candidate since there has been some breaking changes until 1.0.0.

common/app/com/commercetools/sunrise/shoppingcart/CartComponent.java
```diff
@@ -70,16 +79,38 @@
 
     @Override
     public CartCreateCommand onCartCreateCommand(final CartCreateCommand cartCreateCommand) {
+        final CartDraft cartDraft = cartDraftWithAdditionalInfo(cartCreateCommand.getDraft());
```
schleichardt: that example shows we could a wither for the drafts, you can create an issue if you agree https://github.com/commercetools/commercetools-jvm-sdk/issues/new

build.sbt
lauraluiz: No later release found, proposed: https://github.com/commercetools/commercetools-jvm-sdk/issues/1261

common/it/com/commercetools/sunrise/common/TestFixtures.java
lauraluiz: You realize you have been working together with somebody for a long time when you name a function the same way 😆

PR #548, opened 2016-11-24
Add classes, interface and tests related to range facets

jarek170989:
@lauraluiz Could you check if those classes are exactly what we desire? I want to check if everything is fine with work so far to be able to go further. Couple things i've decided to change and i'm not sure if we need some other fields in range object.


move-to-sdk/app/io/sphere/sdk/facets/RangeFacet.java
```diff
@@ -0,0 +1,59 @@
...
+     * Gets the complete options without limitations.
+     * @return the complete possible options
+     */
+    List<FacetOption> getAllOptions();
+
+    /**
+     * Obtains the truncated options list according to the defined limit.
+     * @return the truncated options if the limit is defined, the whole list otherwise
+     */
+    List<FacetOption> getLimitedOptions();
```
lauraluiz: Another things, `limitedOptions` and `allOptions` are necessary concepts for select facets, where you need to somehow control the potentially too large amount of values coming from the platform. But in bucket range facets you know very well the amount, so this mechanism is not useful here.

move-to-sdk/app/io/sphere/sdk/facets/RangeFacet.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    @Nullable
+    Long getThreshold();
+
+    /**
+     * Gets the limit for the maximum amount of options allowed to be displayed in the facet.
+     * @return the limit for the amount of options that can be displayed, or absent if it has no limit
+     */
+    @Nullable
+    Long getLimit();
```
lauraluiz: Limit and threshold are not necessary, for the same reasons as before. The only thing that's needed in terms of controlling whether a facet is displayed or not is what to do when:
- no products contain that attribute (the facet becomes useless), for select facets that was associated with the number of options
- in bucket range facets, whether to display or not the ranges that contain no results.

move-to-sdk/app/io/sphere/sdk/facets/RangeFacetImpl.java
```diff
@@ -0,0 +1,133 @@
...
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import static java.util.stream.Collectors.toList;
+
+public class RangeFacetImpl<T> extends BaseFacet<T> implements RangeFacet<T> {
+
+    private final List<String> selectedValues;
```
lauraluiz: Not sure if `List<String>` is the best way to store these values. In the case of select facet they were a list of values (e.g. red, blue, green), but for slider range facets that would be a single range and for bucket it can be a list of ranges. In any case, I would have here a list of `FacetRange`.

jarek170989:
@lauraluiz Could you look at it? I've changed couple things. I've implemented code only for bucket range just to check if my way of thinking is good.

jarek170989:
@lauraluiz Could you look at it?

move-to-sdk/app/io/sphere/sdk/facets/RangeBucketFacetImpl.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public FacetedSearchExpression<T> getFacetedSearchExpression() {
+        final FacetedSearchExpression<T> facetedSearchExpr;
+        if (selectedRanges.isEmpty()) {
+            facetedSearchExpr = rangeTermFacetedSearchSearchModel.allRanges();
+        } else {
+            facetedSearchExpr = rangeTermFacetedSearchSearchModel.isBetweenAny(
+                    selectedRanges.stream()
+                        .map(facetRange -> FilterRange.of(facetRange.lowerEndpoint().toString(), facetRange.upperEndpoint().toString()))
+                        .collect(toList())
```
lauraluiz: Actually `FilterRange` is the one to use here (as you can notice :P). It has some complexity which I'll try to explain here but I'm always up for a call:
- Filter/FacetRange is internally the same but the allowed ranges are different (that's why we have 2 classes instead of just Range).
- The FacetedSearch model accepts only `FilterRange` because it's NOT well prepared for bucket ranges (that's why we needed to implement this and try). It is assuming you'll always want an initial range of `(* to *)` and thus it only needs to know what are you going to filter by (the filter expression), which is what the `SliderRangeFacet` needs. But the bucket needs to additionally define the initial range (the facet expression). To solve that problem, for now you'll have to create a new `RangeFacetedSearchExpression`, providing the facet and filter expressions yourself:
```
@Override
public FacetedSearchExpression<T> getFacetedSearchExpression() {
    final FacetedSearchExpression<T> facetedSearchExpr;
    if (selectedRanges.isEmpty()) {
        facetedSearchExpr = rangeTermFacetedSearchSearchModel.allRanges();
    } else {
        facetedSearchExpr = rangeTermFacetedSearchSearchModel.isBetweenAny(selectedRanges);
    }
    return buildBucketRangeFacetedSearchExpression(facetedSearchExpr);
}

private RangeFacetedSearchExpression<T> buildBucketRangeFacetedSearchExpression(final FacetedSearchExpression<T> facetedSearchExpr) {
    final List<FacetRange<String>> facetRanges = Collections.emptyList(); // TODO this is an example of facet ranges from bucket range, should be defined in the class
    final RangeTermFacetSearchModel<T, String> searchModel = extractSearchModel(facetedSearchExpr);
    return RangeFacetedSearchExpression.of(searchModel.onlyRange(facetRanges), facetedSearchExpr.filterExpressions());
}

private RangeTermFacetSearchModel<T, String> extractSearchModel(final FacetedSearchExpression<T> facetedSearchExpr) {
    final String attributePath = facetedSearchExpr.facetExpression().attributePath();
    final RangeTermFacetSearchModel<T, String> searchModel = RangeTermFacetSearchModel.of(attributePath, TypeSerializer.ofString());
    return searchModel.withAlias(facetedSearchExpr.facetExpression().alias());
}
```
So in the case of Bucket range facet, you'll need an additional `List<FacetRange>`.

move-to-sdk/app/io/sphere/sdk/facets/RangeBucketFacetImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import static java.util.stream.Collectors.toList;
+
+public class RangeBucketFacetImpl<T> extends BaseFacet<T> implements RangeBucketFacet<T> {
+
+    private final List<FacetRange> selectedRanges;
```
lauraluiz: Same comment as before: don't forget the type parameters, they are important to keep.

move-to-sdk/app/io/sphere/sdk/facets/RangeBucketOption.java
```diff
@@ -0,0 +1,25 @@
...
+
+public interface RangeBucketOption {
+
+    String getLowerEndpoint();
+
+    String getUpperEndpoint();
+
+    Long getCount();
+
+    Boolean getSelected();
```
lauraluiz: `isSelected`

jarek170989:
@lauraluiz I've made changes according to your review. Could you look at it?

lauraluiz:
Now that we have the first approach done I've been pondering about all range facets a bit. I've written down the characteristics of both bucket and slider:

- Does it have an initial range (facet ranges)? Yes -> we send them as facet expressions with our request, No -> we request all ranges.
- If the customer filtered by something, we should have received some ranges to filter our request -> filter ranges
- Can these filter ranges be multiple? Technically yes, with bucket ranges is more evident, but why not with sliders? So a List is required in both
- To display we need max/min/count per range -> that information comes from the request response, except for the max/min in case the initial ranges were defined.

What do you think? If you agree, then I see bucket and slider as the same (simply RangeFacet) and the UI difference coming from the `FacetType` (similar to what we have with SelectFacet, in which case the visual difference comes from the `FacetType` too: Columns, List...).

For the rest, it looks good 👍

jarek170989:
So, i was thinking a lil bit about your remarks:
- i think there isn't an initial range for both types. At the beginning we want to request all ranges, but one thing i'm wondering is that in bucket range we'll have some constant ranges which we can choose and i'm wondering if we should put them into BucketRangeFacet or only in the theme ?
- yes, and i think selectedRanges in BucketRangeFacetImpl and selectedRange in SliderRangeFacetImpl are responsible for that part, right?
- I thought that in slider we can filter only by one range, but technically it can be a multiple, but then in theme there should be slider which can handle choosing more than one range.
- yes, and these information are stored in facetResult in the list of RangeStats, right?

So imo slider and bucket range are similar except the one thing which i've written in first part which i'm not sure about. So it's a good idea to make them as RangeFacet and do it like SelectFacet.

lauraluiz:
- The idea would be to configure the initial ranges that you want (for the bucket) in the `application.conf` and that the Theme is good enough to generate the proper UI given the input data coming from the app. So technically you shouldn't have to touch the Theme for this purpose. When I mean "initial ranges" I refer precisely to these constant ranges from the Bucket. So for Buckets you say "I want Facet Ranges from (x to y) and (z to w)", whereas with Slider you say "I want all Facet Ranges". Then with the RangeStats you can print the ranges in the UI given the min/max and count that you obtain from it.
- Exactly.
- Forget about the Theme, this implementation of Facets doesn't know anything about the Theme, not even about Sunrise (it's located in the `move-to-sdk` module). We have to implement something for all purposes without thinking about what we actually offer in Sunrise. So by definition a slider filter could be multiple, but another story is how the UX designer would think of a suitable UI for that, luckily not our task here ;)
- Yep, we have to extract it from `RangeStats`.

jarek170989:
@lauraluiz I've created one common implementation for bucket and slider. As i understand i can go further with that implementation? Creating tests for it and move to second step?

lauraluiz:
@jarek170989 yep, go ahead 👍

jarek170989:
@lauraluiz I've updated my pr, could you look at it?

lauraluiz:
@jarek170989 still looking good 👍

jarek170989:
@lauraluiz I've made couple changes. Bucket range is finally working well. In project starter in application.conf i've put following temporary configuration of bucket range:
{
    "key": "bucketRange",
    "type": "bucket_range",
    "label": "bucket range",
    "initialRanges": "*-10000,10000-11000,11000-14000,14000-20000,20000-*",
    "expr": "variants.scopedPrice.value.centAmount",
    "count": false,
  }

It's just for testing if everything is working well. I'm not sure how exactly we should provide initial ranges in configuration file. We can easily change it. Also i've created handlebar helper for displaying price properly, because in filter expression for for example 100 euro we need to pass number 10000, but we want to display it as 100. Also in some places i had to remove ending part ".00" of range's endpoints, because sphere response is creating such strings. Could you look at it? Theme for it: https://github.com/commercetools/commercetools-sunrise-theme/pull/438

lauraluiz:
Implemented in v1.0.0-M1

PR #556, opened 2016-12-14
Add provider class for SolutionInfo and file which point at it

jarek170989:
@lauraluiz I'm not sure what values we want to put as "name", "website" or "emergencyContact". I guess it should be hardcoded,right?

### DIFF
--- a/common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
+++ b/common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
@@ -5,9 +5,7 @@ import io.sphere.sdk.meta.BuildInfo;
 
 public class SunriseSolutionInfo extends SolutionInfo {
     public SunriseSolutionInfo() {
-        setName("");
+        setName("commercetools-sunrise-java");
         setVersion(BuildInfo.version());
-        setWebsite("");
-        setEmergencyContact("");
     }
 }
--- a/common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
+++ b/common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
@@ -5,9 +5,9 @@ import io.sphere.sdk.meta.BuildInfo;
 
 public class SunriseSolutionInfo extends SolutionInfo {
     public SunriseSolutionInfo() {
-        setName("JVM-SDK-integration-tests");
+        setName("");
         setVersion(BuildInfo.version());
-        setWebsite("https://github.com/commercetools/commercetools-jvm-sdk");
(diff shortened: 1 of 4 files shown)


lauraluiz:
What happens if you don't define them? I would leave it empty and then in Sunrise Starter Project we would make it easier to add the missing info.

jarek170989:
There won't be any errors. The user agent in corresponding places will have blank text: "commercetools-jvm-sdk/1.4.1 (AHC/2.0) Java/1.8.0_92-b14 (Mac OS X; x86_64) SOLUTION_NAME/SOLUTION_VERSION (+https://website.tld; +info@SOLUTION.com)".  So i'll change it to be empty.

jarek170989:
@lauraluiz I've updated pr.

common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
```diff
@@ -0,0 +1,13 @@
+package com.commercetools.sunrise.common;
+
+import io.sphere.sdk.client.SolutionInfo;
+import io.sphere.sdk.meta.BuildInfo;
+
+public class SunriseSolutionInfo extends SolutionInfo {
+    public SunriseSolutionInfo() {
+        setName("");
+        setVersion(BuildInfo.version());
```
lauraluiz: Have you checked this yields the correct build info for Sunrise? This seems to be coming from the JVM SDK.

common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
```diff
@@ -0,0 +1,13 @@
+package com.commercetools.sunrise.common;
+
+import io.sphere.sdk.client.SolutionInfo;
+import io.sphere.sdk.meta.BuildInfo;
+
+public class SunriseSolutionInfo extends SolutionInfo {
+    public SunriseSolutionInfo() {
+        setName("");
```
lauraluiz: Shouldn't it have `commercetools-sunrise-java`?

common/app/com/commercetools/sunrise/common/SunriseSolutionInfo.java
```diff
@@ -0,0 +1,13 @@
...
+        setVersion(BuildInfo.version());
+        setWebsite("");
+        setEmergencyContact("");
```
lauraluiz: Do we need to explicitly set the website and contact as empty strings? If possible, let's avoid calling the setters if we are not going to set anything.

lauraluiz:
Duplicated https://github.com/commercetools/commercetools-sunrise-java/pull/614

PR #565, opened 2017-02-10
implement observer pattern to refresh category subtree upon global category tree refresh

piobra:
#411 
@lauraluiz please, review

Changes consist of wrapper class that enables thread-safe exchanging of contained category tree.
Observer pattern implementation enables registering and notifying interested listeners (in that case category sub-trees) upon refresh events. The refreshed object notifies its listener which rebuilds it's subtree based on fresh global category tree.

Because of split interfaces projects based on sunrise would need to adjust its guice configuration the same way it was done in this PR.


piobra:
@lauraluiz tests added

lauraluiz:
Had to rework all category tree approach in general and dependant category trees are not singleton anymore: https://github.com/commercetools/commercetools-sunrise-java/pull/590

PR #569, opened 2017-02-21
Travis without docker

katmatt:
Remove docker from our build, switched to container based build environment using trusty distribution which currently provides jdk8u111. 
Added caching of sbt dependencies which hopefully makes our build more stable and faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,6 @@ before_cache:
 
 script:
 - sbt compile test it:test pt:test unidoc
-- ls -l
-- sudo chown -R travis:travis /home/travis
-- ls -l
 - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then ./sbt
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,27 +1,31 @@
...
+  # - https://github.com/travis-ci/travis-ci/issues/3122
+  # - https://github.com/travis-ci/travis-ci/issues/3039
+  branch: md5deep
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot/
+before_cache:
+  # Tricks to avoid unnecessary cache updates
+  - find $HOME/.ivy2 -name "ivydata-*.properties" -delete
+  - find $HOME/.sbt -name "*.lock" -delete
```
lauraluiz: Have you checked that this is still necessary with the `md5deep` branch? It's not clear to me in those issues where the branch is explained whether it solves the lock issue or not.

katmatt: It's still required to improve the stability of the build - especially if a build fails.

PR #582, opened 2017-04-03
Allow to return a subclass of the view models from session

lauraluiz:


lauraluiz:
Not good approach, analyzing other options

PR #586, opened 2017-04-11
Fix logging configuration

lauraluiz:
- No logging is output during tests (except for a single Java money debug message which might be their issue, but for now it's not hurting us).
- Added production and development simplified and optimized logging configuration.


project/TestCommon.scala
```diff
@@ -23,21 +23,26 @@ object TestCommon {
...
+      "org.assertj" % "assertj-core" % "3.6.2" % scopes,
+      "org.mockito" % "mockito-core" % "2.7.9" % scopes
     ),
     dependencyOverrides ++= Set (
       "junit" % "junit" % "4.12" % scopes
     )
   )
 
-  def configJavaWsDependency(scopes: String) = Seq(
+  def configPlayDependencies(scopes: String) = Seq(
```
katmatt: Good idea to rename this method and to add the play-test dependency here👍

PR #590, opened 2017-04-21
Category tree improvements, closes #411, closes #218, closes #217, closes #419, closes #407, closes #351

### LINKED ISSUE #217: Implement dynamic New category
Custom attribute with ID of the referenced category, then programmatically filter by some time limit (configurable) and this category ID.

It is not possible to use ext. ID since it is not possible to filter by category ext ID.

### LINKED ISSUE #218: Dynamic sales category
Same procedure as "New" category https://github.com/sphereio/sphere-sunrise/issues/217 but filtering by whether the product has discount or not.

### LINKED ISSUE #351: Hide empty categories
A common feature request is that the storefront must never show categories that contain no products (unless they are "dynamic" like sales). It's not about manually activating / deactivating categories but about an additional automatism. 

The core API does not directly support the case, but it would be great if the SUNRISE implementation would provide a best practice suggestion on how to implement the requirement (nearly all projects will have to cover it some way). 

A relatively "cheap" way to find out could be a facets only search on the category ID.  But the limitation in number of facets returned is a bit unclear and intransparent.

### LINKED ISSUE #407: Job for Category and ProductType autorefresh
@butenkor suggested that it would be nice to set a intervall in application.conf when the shop should refresh the categories and product types

So it would be like a cronjob.

### LINKED ISSUE #411: @Named dependency injected categories should get refreshed when the whole category tree gets refreshed

### LINKED ISSUE #419: support category tree with with certain root
Currently just all categories are used starting from the roots, actually it would be nice if we could use a root category "web" which contains the categories shown in the web shop so we still can build other category trees.

lauraluiz:
#351 can be improved once backend enables filtering by productCount.


lauraluiz:
Have to figure out why the IT fail in CI

common/app/com/commercetools/sunrise/ctp/graphql/GraphQLException.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class GraphQLException extends RuntimeException  {
+
+    private static final long serialVersionUID = 0L;
+
+    public GraphQLException(final String message, final Throwable cause) {
+        super(message, cause);
+    }
+
+    public GraphQLException(final String message, final HttpResponse httpResponse) {
```
katmatt: Why do we need this constructor? It doesn't use the 2nd parameter...

common/app/com/commercetools/sunrise/ctp/graphql/GraphQLRequest.java
```diff
@@ -0,0 +1,14 @@
+package com.commercetools.sunrise.ctp.graphql;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.sphere.sdk.client.SphereRequest;
+
+import javax.annotation.Nullable;
+
+public interface GraphQLRequest<T> extends SphereRequest<T> {
```
katmatt: I think for now it's a good idea to extend from `SphereRequest`, but I don't know if the type parameter `T`will make sense for GraphQL in the future. But I really like that we now use our GraphQL API in sunrise 👍 and it will give us more insights on how to support it in our SDK 😉

common/app/com/commercetools/sunrise/ctp/graphql/SingleGraphQLRequest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public HttpRequestIntent httpRequestIntent() {
+        final JsonNode variables = getVariables();
+        final String variablesPart = variables == null ? null : String.format("\"variables\": %s", variables);
+        final String queryPart = String.format("\"query\": \"%s\"", getQuery().replace("\n", "\\n").replace("\"", "\\\""));
+        final String body = Stream.of(queryPart, variablesPart)
+                .filter(Objects::nonNull)
+                .collect(joining(",", "{", "}"));
```
katmatt: Cool 👍 This is a very good example which shows the power of the Stream API 👍

common/app/com/commercetools/sunrise/ctp/graphql/SingleGraphQLRequest.java
```diff
@@ -0,0 +1,53 @@
...
+            }
+        } else {
+            throw new GraphQLException("No data found in response", httpResponse);
+        }
+    }
+
+    @Override
+    public HttpRequestIntent httpRequestIntent() {
+        final JsonNode variables = getVariables();
+        final String variablesPart = variables == null ? null : String.format("\"variables\": %s", variables);
```
katmatt: According to the javadoc [JsonNode#tostring()](http://fasterxml.github.io/jackson-databind/javadoc/2.8/com/fasterxml/jackson/databind/JsonNode.html#toString()) we shouldn't rely on `JsonNode#toString` method here. Instead we should use one of the helper methods of `SphereJsonUtils` here.

common/app/com/commercetools/sunrise/framework/viewmodels/forms/FormSettingsWithDefault.java
```diff
@@ -0,0 +1,21 @@
...
+
+    T getDefaultValue();
+
+    /**
+     * Finds one selected valid value for this form in the HTTP request.
+     * If many valid values are selected, which one is going to be returned is non-deterministic.
+     * @param httpContext current HTTP context
+     * @return a valid selected value for this form, or the default value if no valid value is selected
+     */
+    default T getSelectedValueOrDefault(final Http.Context httpContext) {
```
katmatt: I really like how you changed the logic here: use `Optional<T>` when there is no default and use `T`if there is a default value 👍

common/test/com/commercetools/sunrise/framework/viewmodels/forms/FormSettingsWithDefaultTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.commercetools.sunrise.framework.viewmodels.forms.FormTestUtils.someQueryString;
+import static com.commercetools.sunrise.framework.viewmodels.forms.FormTestUtils.testWithHttpContext;
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FormSettingsWithDefaultTest {
```
katmatt: I like unit tests 👍

common/it/com/commercetools/sunrise/categorytree/CategoriesWithProductCountQueryIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.Test;
+
+import java.util.List;
+import java.util.function.Consumer;
+
+import static com.commercetools.sunrise.it.CategoryTestFixtures.categoryDraft;
+import static com.commercetools.sunrise.it.CategoryTestFixtures.withCategory;
+import static com.commercetools.sunrise.it.ProductTestFixtures.*;
+import static com.commercetools.sunrise.it.ProductTypeTestFixtures.productTypeDraft;
+import static com.commercetools.sunrise.it.ProductTypeTestFixtures.withProductType;
```
katmatt: Having a public sdk test lib would really help here... But great that you created an integration test 👍

lauraluiz:
@katmatt issue of HTTP Response handled in exception and properly transform JSON into String fixed! :)

Issue #395, opened 2016-05-18
MyAccount - Change password


mmoelli:

fhaertig:
Just to check: Am I correct that there is already a template file for my-account-change-password, but the controller is missing and the HAL link is not present in PageMeta?

lauraluiz:
@fhaertig yes, there is actually a template file for that, but we are missing all backend side (controllers, HAL Link, routes, etc.).

selo-ecube:
Do you have plans to implement _change password_ (and also _reset password_) in the coming weeks? Or do you solely focus on stabilizing the framework instead?

lauraluiz:
@selo-ecube second option: stabilizing the framework. Although once the backend code is stable we will work on stabilizing the Theme and at the same time we will implement some backend functionality of the wishlist. But features that require both frontend and backend won't be in the roadmap until v1.0.0 is reached.

PR #606, opened 2017-05-16
Change password

piobra:
@katmatt please, review
#395

### DIFF
--- a/common/app/com/commercetools/sunrise/framework/reverserouters/myaccount/authentication/AuthenticationLinksControllerComponent.java
+++ b/common/app/com/commercetools/sunrise/framework/reverserouters/myaccount/authentication/AuthenticationLinksControllerComponent.java
@@ -26,7 +26,6 @@ public class AuthenticationLinksControllerComponent extends AbstractLinksControl
         meta.addHalLink(reverseRouter.logOutProcessCall(), "logOut");
         meta.addHalLink(reverseRouter.logInPageCall(), "signUp");
         meta.addHalLink(reverseRouter.signUpProcessCall(), "signUpSubmit");
-        meta.addHalLink(reverseRouter.changePasswordPageCall(), "changePassword");
         meta.addHalLink(reverseRouter.changePasswordProcessCall(), "changePassword");
     }
 }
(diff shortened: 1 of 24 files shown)


my-account/app/com/commercetools/sunrise/myaccount/authentication/changepassword/viemodels/ChangePasswordPageContentFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    protected void initialize(final ChangePasswordPageContent viewModel, final Void input, final Form<? extends ChangePasswordFormData> form) {
+        super.initialize(viewModel, input, form);
+        fillChangePasswordForm(viewModel, form);
+    }
+
+    @Override
+    protected void fillTitle(final ChangePasswordPageContent viewModel, final Void input, final Form<? extends ChangePasswordFormData> form) {
+        viewModel.setTitle(pageTitleResolver.getOrEmpty("my-account:changePassword.title")); // TODO messages missing in FE templates
```
katmatt: We have a translation for the title in our [my-account.yaml](https://github.com/commercetools/commercetools-sunrise-theme/blob/master/input/i18n/en/my-account.yaml#L86), but it's called `changePasswordTitle`. We should later rename it to `title`, but for now I would suggest to use `my-account:changePassword.changePasswordTitle` here.

common/app/com/commercetools/sunrise/framework/reverserouters/myaccount/authentication/AuthenticationLinksControllerComponent.java
```diff
@@ -26,5 +26,7 @@ protected void addLinksToPage(final PageMeta meta, final AuthenticationReverseRo
         meta.addHalLink(reverseRouter.logOutProcessCall(), "logOut");
         meta.addHalLink(reverseRouter.logInPageCall(), "signUp");
         meta.addHalLink(reverseRouter.signUpProcessCall(), "signUpSubmit");
+        meta.addHalLink(reverseRouter.changePasswordPageCall(), "changePassword");
```
katmatt: Additionally I would suggest to introduce a separate `ChangePasswordLinksControllerComponent` class and a corresponding `ChangePasswordReverseRouter` interface with a `DefaultChangePasswordReverseRouter` class.

piobra:
@katmatt thanks for remarks.
I made adjustments according to them.
Regarding routes I changed to "changepassword" rather than "password" because there will be also reset password routes. I'm currently working on them. I'm thinking about making them "resetpassword".

If I did it your way we'd end up with sth like "password" and "reset" which in my opinion is not clear and self descriptive.

katmatt:
@piobra I see why you choose to use `changepassword`, but I still would suggest to use `password`and later to add `password/reset` for reseting the password.

With regards to working on the reset/recover password feature: I think that this feature requires more preparation (e.g. sending a confirmation link etc.) and I don't think that it makes sense to work on it now. And that's what we already discussed with Florian.

piobra:
@katmatt done

PR #607, opened 2017-05-29
Wishlist

### LINKED ISSUE #334: Wishlist

katmatt:
Closes #334


katmatt:
@lauraluiz Using `Wishlist` when the type is a `ShoppingList` didn't feel right, so let's discuss if we both can come up with better names 😉

lauraluiz:
My quick thoughts on that (without checking the code):
Shopping list is the general term for all kind of lists: wishlist, wedding list or customized lists (e.g. "review later" list).
With that being said, it depends on the solution, whether it's too "wishlist" focused or not.
And now I check the code ;)

my-account/app/com/commercetools/sunrise/myaccount/wishlist/WishlistCreatorBySession.java
```diff
@@ -0,0 +1,40 @@
...
+        super(sphereClient, hookRunner);
+        this.customerInSession = customerInSession;
+    }
+
+    public CompletionStage<ShoppingList> get() {
+        final Reference<Customer> customer = customerInSession.findCustomerId()
+                .map(Customer::referenceOfId)
+                .orElse(null);
+
+        final ShoppingListDraftDsl wishlist = ShoppingListDraftBuilder.of(LocalizedString.ofEnglish("Wishlist"))
```
lauraluiz: So to summarize, we are missing the `AbstractWishlistCreateExecutor` (check https://github.com/commercetools/commercetools-sunrise-java/blob/master/shopping-cart/app/com/commercetools/sunrise/framework/cart/addlineitem/AbstractCartCreateExecutor.java) and a `buildRequest` protected method that developers can modify to add simple things like expansion paths.

app/controllers/myaccount/AddToWishlistController.java
```diff
@@ -0,0 +1,24 @@
...
+
+import javax.inject.Inject;
+
+public class AddToWishlistController extends SunriseAddToWishlistController {
+
+    @Inject
+    public AddToWishlistController(final ContentRenderer contentRenderer,
+                                   final FormFactory formFactory,
+                                   final WishlistLineItemFormData formData,
+                                   final WishlistFinderBySession wishlistFinder,
```
lauraluiz: This should be injecting the interface, not the concrete method.

my-account/app/com/commercetools/sunrise/myaccount/wishlist/DefaultAddToWishlistControllerAction.java
```diff
@@ -0,0 +1,39 @@
...
+    protected DefaultAddToWishlistControllerAction(final SphereClient sphereClient, final HookRunner hookRunner) {
+        super(sphereClient, hookRunner);
+    }
+
+    @Override
+    public CompletionStage<ShoppingList> apply(final ShoppingList shoppingList, final WishlistLineItemFormData addToWishlistFormData) {
+        return executeRequest(shoppingList, buildRequest(shoppingList, addToWishlistFormData));
+    }
+
+    private ShoppingListUpdateCommand buildRequest(final ShoppingList shoppingList, final WishlistLineItemFormData addToWishlistFormData) {
```
lauraluiz: Applies for the rest of controller actions too.

my-account/app/com/commercetools/sunrise/myaccount/wishlist/SunriseWishlistController.java
```diff
@@ -0,0 +1,86 @@
...
+        this.controllerAction = controllerAction;
+        this.wishlistPageContentFactory = wishlistPageContentFactory;
+        this.paginationSettings = paginationSettings;
+        this.entriesPerPageFormSettings = entriesPerPageFormSettings;
+    }
+
+    @EnableHooks
+    @SunriseRoute(WishlistReverseRouter.WISHLIST_PAGE_CALL)
+    public CompletionStage<Result> show(final String languageTag) {
+        return wishlistFinder.getOrCreate()
```
lauraluiz: Even though this solution is more direct, it has a big drawback compared to having the finder only finding and the creator only creating: you are creating a shopping list every time somebody accesses his/hers. And it happens the same in the remove line item or clear actions.
As with carts, we have to be careful not to create unnecessary resources. So only when an item is added to the wishlist, this should be created if it doesn't yet exist.
Why not use the same system as used in the rest of the controllers of having the interface `WithRequiredWishlist` (e.g. https://github.com/commercetools/commercetools-sunrise-java/blob/master/shopping-cart/app/com/commercetools/sunrise/framework/WithRequiredCart.java), which can be reused for other situations and forces to define a method to handle not found situations. In the case of the add line item, this situation will call the `WishlistCreator` and try the `process` method again.
Before we had it like you just did, but then devs had to copy paste the contents of this method and couldn't decide what to do in case of not finding it. Now they know they can always control it under the `handleNotFoundX` :)
But let me know if you see problems about it that I fail to see or if you have better suggestions.

my-account/app/com/commercetools/sunrise/myaccount/wishlist/SunriseWishlistController.java
```diff
@@ -0,0 +1,86 @@
...
+                .thenComposeAsync(this::getWishlist, HttpExecution.defaultContext())
+                .thenComposeAsync(this::showPage, HttpExecution.defaultContext());
+    }
+
+    @EnableHooks
+    @SunriseRoute(WishlistReverseRouter.CLEAR_WISHLIST_PROCESS)
+    public CompletionStage<Result> clear(final String languageTag) {
```
lauraluiz: I would move the clear to its own controller. Why having it here while the other actions have their own? Don't treat poor clear as a second-class citizen in Sunrise 😉 *joke*

my-account/app/com/commercetools/sunrise/myaccount/wishlist/SunriseRemoveFromWishlistController.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class SunriseRemoveFromWishlistController extends SunriseContentFormController
+        implements WithFormFlow<ShoppingList, ShoppingList, WishlistLineItemFormData> {
+    private final WishlistLineItemFormData formData;
+    private final WishlistFinderBySession wishlistFinder;
+    private final RemoveFromWishlistControllerAction controllerAction;
+    private final WishlistReverseRouter reverseRouter;
+
+    @Inject
+    protected SunriseRemoveFromWishlistController(final ContentRenderer contentRenderer, final FormFactory formFactory, final WishlistLineItemFormData formData,
```
lauraluiz: Just leaving this comment here so I don't forget to think about whether it's better to have no associated template and rely on redirects for handling bad requests. As long as error messages appear the same it can certainly be better, and they rely on flash so it should work. But need to check.

katmatt:
@lauraluiz I'm now done with the changes. I'm looking forward to your review 😄

common/app/com/commercetools/sunrise/sessions/wishlist/DefaultWishlistInSession.java
```diff
@@ -33,17 +41,24 @@ public DefaultWishlistInSession(final Configuration configuration, final ObjectS
...
     protected Logger getLogger() {
         return LOGGER;
     }
 
     @Override
-    protected void storeAssociatedData(@NotNull final ShoppingList value) {
-        session.overwriteValueByKey(wishlistIdSessionKey, value.getId());
+    protected void storeAssociatedData(@NotNull final ShoppingList wishlist) {
+        session.overwriteValueByKey(wishlistIdSessionKey, wishlist.getId());
+        session.overwriteObjectByKey(wishlistSessionKey, wishlistViewModelFactory.create(wishlist));
```
lauraluiz: Having the mini-wishlist and mini-cart in the session together won't work anymore, even with the product limit. It's been some time since I wanted to use cache by default, because everybody ends up with the same space issue. If we merge this PR we will need to address the session issue before releasing...

wishlist/app/com/commercetools/sunrise/wishlist/WishlistFinderBySession.java
```diff
@@ -0,0 +1,61 @@
...
+import io.sphere.sdk.shoppinglists.ShoppingList;
+import io.sphere.sdk.shoppinglists.queries.ShoppingListQuery;
+
+import javax.inject.Inject;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+
+public class WishlistFinderBySession extends AbstractSingleShoppingListQueryExecutor implements WishlistFinder {
```
lauraluiz: This class doesn't provide any methods to override (only get, which is the same as implementing your own class from scratch) so I would make it package scope and final.

wishlist/app/com/commercetools/sunrise/wishlist/controllers/DefaultRemoveFromWishlistControllerAction.java
```diff
@@ -0,0 +1,34 @@
...
+        return executeRequest(wishlist, buildRequest(wishlist, removeWishlistFormData));
+    }
+
+    protected ShoppingListUpdateCommand buildRequest(final ShoppingList wishlist, final RemoveWishlistLineItemFormData removeWishlistFormData) {
+        final LineItem lineItemToRemove = wishlist.getLineItems().stream()
+                .filter(lineItem -> removeWishlistFormData.lineItemId().equals(lineItem.getId()))
+                .findFirst()
+                .get();
+
+        final RemoveLineItem removeLineItem = RemoveLineItem.of(lineItemToRemove);
```
lauraluiz: Why using streams to find the line item? RemoveLineItem accepts a lineItemId directly.

wishlist/app/com/commercetools/sunrise/wishlist/controllers/ClearWishlistControllerAction.java
```diff
@@ -9,6 +9,13 @@
 
 @ImplementedBy(DefaultClearWishlistControllerAction.class)
 public interface ClearWishlistControllerAction extends ControllerAction, Function<ShoppingList, CompletionStage<ShoppingList>> {
+    /**
+     * Removes all line items from the given wishlist.
+     *
+     * @param wishlist the wishlist
+     *
+     * @return the completion stage for this action
+     */
```
lauraluiz: Maybe it's better to have a class javadoc than having it for the method, as in this case (as in most cases) `@param` and `@return` documentation are too verbose and provide no additional info. That, or be more specific with the description:
```
    * @param wishlist the {@link shoppingList} from which to remove all line items
    *
    * @return the completion stage for the updated wishlist with no items.
```

common/app/com/commercetools/sunrise/sessions/wishlist/DefaultWishlistInSession.java
katmatt: @lauraluiz Why wouldn't they work anymore? Because we will exceed the memory limit of the session?

wishlist/app/com/commercetools/sunrise/wishlist/controllers/DefaultRemoveFromWishlistControllerAction.java
katmatt: Good catch and you're right 👍

wishlist/app/com/commercetools/sunrise/wishlist/WishlistFinderBySession.java
katmatt: Good point and I agree 👍

common/app/com/commercetools/sunrise/sessions/wishlist/DefaultWishlistInSession.java
lauraluiz: Yes, it has nothing to do with wishlist, but before we had a limit of 5-6 items until the session stopped working (and mini-cart was not updated). but if we add wishlist plus the additional information, it will not work with more than 2 items on average. Since I enabled cache in categoryTree, I've considered moving fully to cache. It's actually quite easy, just changing the default implementation of the ObjectStoringStrategy class. But we have to keep it in mind :)

lauraluiz: And specially document it, of course. It will work with one application, but on multi-servers it will fail miserably if not used a more advanced caching system (unless configured properly).

common/app/com/commercetools/sunrise/framework/viewmodels/content/PageContent.java
```diff
@@ -36,4 +38,13 @@ public void addComponent(final ViewModelComponent component) {
...
         this.components.add(component);
     }
+
+    public WishlistViewModel getWishlist() {
+        return wishlist;
+    }
+
+    public void setWishlist(final WishlistViewModel wishlist) {
+        this.wishlist = wishlist;
+    }
```
katmatt: Using `extendedViewModel` worked fine and as we discussed - should be fine for now.

wishlist/app/com/commercetools/sunrise/wishlist/viewmodels/WishlistPageContentFactory.java
```diff
@@ -20,52 +15,56 @@
...
+    @Override
+    protected final void initialize(final WishlistPageContent viewModel, final ShoppingList input) {
         super.initialize(viewModel, input);
 
-        fillProducts(viewModel, input);
-        fillItemsInTotal(viewModel, input);
+        if (input != null) {
+            fillProducts(viewModel, input);
+            fillItemsInTotal(viewModel, input);
+        }
```
katmatt: I fully agree 👍

wishlist/app/com/commercetools/sunrise/wishlist/controllers/SunriseWishlistController.java
```diff
@@ -0,0 +1,62 @@
...
+    @SunriseRoute(WishlistReverseRouter.WISHLIST_PAGE)
+    public CompletionStage<Result> show(final String languageTag) {
+        return requireWishlist(this::showPage);
+    }
+
+    @Nullable
+    @Override
+    public String getTemplateName() {
+        return "my-account-wishlist";
+    }
```
katmatt: 👍

wishlist/app/com/commercetools/sunrise/wishlist/controllers/SunriseClearWishlistController.java
```diff
@@ -0,0 +1,59 @@
...
+    public CompletionStage<Result> process(final String languageTag) {
+        return requireWishlist(this::processRequest);
+    }
+
+    @Override
+    public CompletionStage<ShoppingList> executeAction(final ShoppingList wishlist) {
+        return controllerAction.apply(wishlist);
+    }
+
+    public abstract CompletionStage<Result> handleSuccessfulAction(final ShoppingList output);
```
katmatt: 👍

wishlist/app/com/commercetools/sunrise/wishlist/controllers/SunriseClearWishlistController.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public WishlistFinder getWishlistFinder() {
+        return wishlistFinder;
+    }
+
+    @Override
+    public CompletionStage<Result> handleClientErrorFailedAction(final ShoppingList input, final ClientErrorException clientErrorException) {
+        // No CTP call involved, this should never be called
```
katmatt: I removed the confusing comment 👍

wishlist/app/com/commercetools/sunrise/wishlist/controllers/ClearWishlistControllerAction.java
katmatt: As discussed, I would still use javdoc on the method because it provides more context. But I agree with your improved method javadoc and applied this style to all wishlist controller actions 👍

katmatt:
@lauraluiz I applied all requested/discussed changes 😄 Looking forward to your next review 😉

common/app/com/commercetools/sunrise/framework/viewmodels/content/wishlist/WishlistViewModelFactory.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    protected final void initialize(final WishlistViewModel viewModel, final ShoppingList wishlist) {
+        if (wishlist != null) {
+            fillTitle(viewModel, wishlist);
+
+            final List<LineItem> lineItems = wishlist.getLineItems();
+
+            fillQuantity(viewModel, lineItems);
+            fillLineItems(viewModel, lineItems);
+        }
```
lauraluiz: There's still this null check in the initialize method ;)

lauraluiz: And pass the whole wishlist to those methods, we should not make assumptions of what the user may need to fill those fields, so it's better to pass exactly the same everywhere.

common/app/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactory.java
```diff
@@ -121,11 +125,12 @@ private long calculateTopThresholdPage(final long currentPage, final long totalP
...
-        if (pagedResult.isLast()) {
-            return currentPage;
-        } else if (pagedResult.getCount() > 0) {
-            final Double totalPages = Math.ceil(pagedResult.getTotal() / pagedResult.getCount());
+    private long calculateTotalPages(final PagedResult<?> pagedResult) {
+        final Long count = pagedResult.getCount();
+        if (count > 0) {
+            final Long total = pagedResult.getTotal();
+            final Http.Context context = Http.Context.current();
+            final Double totalPages = Math.ceil(((double) total) / entriesPerPageFormSettings.getLimit(context));
```
lauraluiz: You didn't answer to this yet :P

common/app/com/commercetools/sunrise/sessions/wishlist/DefaultWishlistInSession.java
```diff
@@ -0,0 +1,64 @@
...
+    private static final String DEFAULT_WISHLIST_ID_SESSION_KEY = "sunrise-wishlist-id";
+    private static final String DEFAULT_WISHLIST_SESSION_KEY = "sunrise-wishlist";
+
+    private final String wishlistIdSessionKey;
+    private final String wishlistSessionKey;
+    private final ObjectStoringSessionStrategy session;
+    private final WishlistViewModelFactory wishlistViewModelFactory;
+
+    @Inject
+    public DefaultWishlistInSession(final Configuration configuration, final ObjectStoringSessionStrategy session,
```
lauraluiz: Injected constructor -> `protected`

wishlist/app/com/commercetools/sunrise/wishlist/WishlistCreator.java
```diff
@@ -0,0 +1,17 @@
+package com.commercetools.sunrise.wishlist;
+
+import com.commercetools.sunrise.wishlist.controllers.DefaultWishlistCreator;
+import com.google.inject.ImplementedBy;
+import io.sphere.sdk.shoppinglists.ShoppingList;
+
+import java.util.concurrent.CompletionStage;
+
+@ImplementedBy(DefaultWishlistCreator.class)
+public interface WishlistCreator {
```
lauraluiz: Could you tag it as a `@FunctionalInterface` and extend it from `ResourceFinder` and `Supplier<CompletionStage<ShoppingList>>` (this one is just to be coherent with the naming, I might remove it in the future though it doesn't hurt)? That's how all other creators/finders look like.

wishlist/app/com/commercetools/sunrise/wishlist/WishlistFinder.java
```diff
@@ -0,0 +1,18 @@
+package com.commercetools.sunrise.wishlist;
+
+import com.google.inject.ImplementedBy;
+import io.sphere.sdk.shoppinglists.ShoppingList;
+
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+@ImplementedBy(WishlistFinderBySession.class)
+public interface WishlistFinder {
```
lauraluiz: Same as with WishlistCreator -> functionalinterface, resourcefinder, supplier.

wishlist/app/com/commercetools/sunrise/wishlist/controllers/AddToWishlistControllerAction.java
```diff
@@ -0,0 +1,23 @@
...
+import com.commercetools.sunrise.framework.controllers.ControllerAction;
+import com.commercetools.sunrise.wishlist.viewmodels.AddWishlistLineItemFormData;
+import com.google.inject.ImplementedBy;
+import io.sphere.sdk.shoppinglists.ShoppingList;
+
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+
+@ImplementedBy(DefaultAddToWishlistControllerAction.class)
+public interface AddToWishlistControllerAction extends ControllerAction, BiFunction<ShoppingList, AddWishlistLineItemFormData, CompletionStage<ShoppingList>> {
```
lauraluiz: `@FunctionalInterface`

wishlist/app/com/commercetools/sunrise/wishlist/controllers/ClearWishlistControllerAction.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.commercetools.sunrise.framework.controllers.ControllerAction;
+import com.google.inject.ImplementedBy;
+import io.sphere.sdk.shoppinglists.ShoppingList;
+
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+@ImplementedBy(DefaultClearWishlistControllerAction.class)
+public interface ClearWishlistControllerAction extends ControllerAction, Function<ShoppingList, CompletionStage<ShoppingList>> {
```
lauraluiz: `@FunctionalInterface`

wishlist/app/com/commercetools/sunrise/wishlist/controllers/RemoveFromWishlistControllerAction.java
```diff
@@ -0,0 +1,23 @@
...
+import com.commercetools.sunrise.framework.controllers.ControllerAction;
+import com.commercetools.sunrise.wishlist.viewmodels.RemoveWishlistLineItemFormData;
+import com.google.inject.ImplementedBy;
+import io.sphere.sdk.shoppinglists.ShoppingList;
+
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+
+@ImplementedBy(DefaultRemoveFromWishlistControllerAction.class)
+public interface RemoveFromWishlistControllerAction extends ControllerAction, BiFunction<ShoppingList, RemoveWishlistLineItemFormData, CompletionStage<ShoppingList>> {
```
lauraluiz: `@FunctionalInterface`

wishlist/app/com/commercetools/sunrise/wishlist/controllers/SunriseWishlistController.java
```diff
@@ -0,0 +1,55 @@
...
+import io.sphere.sdk.shoppinglists.ShoppingList;
+import play.mvc.Result;
+
+import javax.inject.Inject;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * This controller is used to view the current wishlist.
+ */
+public class SunriseWishlistController extends SunriseContentController implements WithQueryFlow<ShoppingList>, WithRequiredWishlist {
```
lauraluiz: It should be `abstract` (even though it has no abstract method left to be implemented, it won't work by default without an associated templateName, so in general it should not be used directly).

common/app/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactory.java
katmatt: This class had some bugs which I found when i added the pagination for the wishlist. But since we decided to remove the pagination for the wishlist, I can also revert these changes.

lauraluiz: If there are bugs, we have to fix them 😄 
Could you write (here or in another branch and we revert these) a UT that shows the bugs that were fixed?

katmatt: Yes, I will check the existing UT to make sure that it covers the bug that I found 👍

katmatt:
@lauraluiz I applied the changes we discussed and added the improved pagination test. I'm looking forward to your code review 😉

katmatt:
I created a PR for sunrise-theme, when you approved the PR and the webjar is published I will then remove the templates from this PR :-)

PR #614, opened 2017-06-07
Add provider class for solutioninfo and file which point at it

acbeni:


common/it/com/commercetools/sunrise/common/version/UserAgentIntegrationTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+
+    @Test
+    public void userAgent() throws Exception {
+        final SphereClientConfig config = sphereClientConfig();
+        try (final FakeHttpClient httpClient = new FakeHttpClient()) {
+            try (final SphereClient client = SphereClientFactory.of(() -> httpClient).createClient(config)) {
+                client.execute(ProjectGet.of()).toCompletableFuture().join();
+                assertThat(httpClient.getLastUserAgent())
+                        .contains("sunrise-java-shop-framework");
```
lauraluiz: It's a pity we cannot check whether the version is there. Maybe also that the JVM SDK part is included too.

Having a user-agent that looks like this:
```
commercetools-jvm-sdk/1.13.0 (AHC/2.0) Java/1.8.0_92-b14 (Mac OS X; x86_64) sunrise-java-shop-framework/1.0.0-M4-SNAPSHOT
```
We could maybe check `commercetools-jvm-sdk` is there and `sunrise-java-shop-framework` as we have it, and with the thing I explain in the next comment we could even check the version. WDYT?

project/Version.scala
```diff
@@ -30,4 +30,26 @@ object Version {
...
+          |    public SunriseSolutionInfo() {
+          |        setName("sunrise-java-shop-framework");
+          |        setVersion("${version.value}");
+          |    }
+          |
+          |}
+          |""".stripMargin)
+      Seq(file)
+    }.taskValue)
+
```
lauraluiz: It's good that you found a viable solution 👍 
I found a possible option that would make the code nicer, using an SBT plugin that allows to access SBT vars from the source code: https://github.com/sbt/sbt-buildinfo
That would great if it works well, and it would even allow us to test everything. Looks promising, don't you think? :)

test-lib/app/com/commercetools/sunrise/it/WithSphereClient.java
```diff
@@ -54,4 +59,10 @@ protected static SphereClientConfig sphereClientConfig() {
         final String clientSecret = configuration.getString("ctp.it.clientSecret");
         return SphereClientConfig.of(projectKey, clientId, clientSecret);
     }
+
+    protected static HttpClient newHttpClient() {
```
lauraluiz: Why do we need this method here? If we need it for a particular test, I would leave it on that place. I wouldn't polute this class which is meant as a helper class for developers.

lauraluiz:
Ah I have an idea what the cause can be of the failed IT, I'm checking.

test-lib/app/com/commercetools/sunrise/it/WithSphereClient.java
lauraluiz: Ignore this, I've already fixed it :)

common/it/com/commercetools/sunrise/common/version/UserAgentIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
+package com.commercetools.sunrise.common.version;
```
lauraluiz: I've missed that. This package is only for the controller that provides the version of the app via an endpoint. The solution info sent via user-agent is not related with that controller, so I would place the classes somewhere else, for example in `com.commercetools.sunrise.ctp.useragent`

common/it/com/commercetools/sunrise/common/version/UserAgentIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.Test;
+
+import javax.annotation.Nullable;
+import java.util.concurrent.CompletionStage;
+
+import static com.commercetools.sunrise.it.SphereClientFixtures.provideHttpClient;
+import static com.commercetools.sunrise.it.SphereClientFixtures.provideSphereClient;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class UserAgentIntegrationTest {
```
lauraluiz: And I've missed this too 😨 
I guess I was focusing too much in the code :P
So the test class should always have the same base name as the class it's testing (that helps a lot maintenance since you can quickly associate both classes). In this case `SunriseSolutionInfoIntegrationTest`.

project/Version.scala
acbeni: Actually this plugin generate scala code, but i would agree the way i did it wouldn't let us now if some error happened during the build process thats why i will build a class BuildInfo in using the same way i generated SunriseVersionInfo.java, the only diff is that i'm gonna refernece this class in SunriseSolutionInfo, so if an error happens the code wont compile

common/it/com/commercetools/sunrise/common/version/UserAgentIntegrationTest.java
acbeni: +! totally agree

lauraluiz:
We should make sure the tests are not flaky as they seem they are before merging. So far it's failing almost 50% of the times, which would become hell for future PR :(
But this topic is not urgent at all, so take your time.

common/it/com/commercetools/sunrise/common/version/UserAgentIntegrationTest.java
lauraluiz: @acbeni now you should be able to do this, right?

katmatt:
@lauraluiz Looks like the last builds where quite stable 😉 How about merging this PR?

lauraluiz:
Sounds good, but I'd rather leave it for the next release and review it with calm ;)

PR #616, opened 2017-06-09
Fix payment not expanded error message, closes #502

### LINKED ISSUE #502: Fix payment not expanded error message
PaymentInfo is null when there's no payment yet. The error should be logged only when the references for the list of payments are not expanded.

``` java
final List<Reference<Payment>> paymentRefs = Optional.ofNullable(cart.getPaymentInfo())
        .map(PaymentInfo::getPayments)
        .orElseGet(() -> {
            logger.error("Payment info is not expanded in cart: the new payment information can be saved but the previous payments will not be removed.");
            return emptyList();
        });
```

lauraluiz:
Before it was logging the warning when paymentInfo was null (which happens when there are no payments yet) and now only when there are really not expanded. Also added a bit of documentation to let it clear.
I was thinking on fetching the payments if they are not expanded, but I'm afraid that will make people choose the easiest path, which affects a lot performance.

### DIFF
--- a/shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
+++ b/shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
@@ -3,6 +3,7 @@ package com.commercetools.sunrise.framework.checkout.payment;
 import com.commercetools.sunrise.framework.AbstractCartUpdateExecutor;
 import com.commercetools.sunrise.framework.hooks.HookRunner;
 import io.sphere.sdk.carts.Cart;
+import io.sphere.sdk.carts.PaymentInfo;
 import io.sphere.sdk.carts.commands.CartUpdateCommand;
 import io.sphere.sdk.carts.commands.updateactions.AddPayment;
 import io.sphere.sdk.carts.commands.updateactions.RemovePayment;
@@ -67,10 +68,8 @@ public class DefaultCheckoutPaymentControllerAction extends AbstractCartUpdateEx
     private List<Payment> findPaymentsToRemove(final Cart cart, final Payment payment) {
         return Optional.ofNullable(cart.getPaymentInfo())
                 .map(paymentInfo -> {
-                    final boolean referenceAreExpanded = paymentInfo.getPayments().stream()
-                            .allMatch(reference -> reference.getObj() != null);
-                    if (!referenceAreExpanded) {
-                        LOGGER.warn("Payment info is not expanded in cart: the new payment information can be saved but the previous payments will not be removed.");
+                    if (!hasReferencesExpanded(paymentInfo)) {
+                        LOGGER.warn("Payments are not expanded in cart: the new payment information can be saved but the previous payments will not be removed.");
                     }
                     return paymentInfo.getPayments().stream()
                             .filter(paymentRef -> !paymentRef.referencesSameResource(payment))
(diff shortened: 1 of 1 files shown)


shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
```diff
@@ -102,4 +103,9 @@ private PaymentMethodInfo findPaymentMethod(final List<PaymentMethodInfo> paymen
                 .findAny()
                 .orElseThrow(() -> new RuntimeException("No valid payment method found")); // Should not happen after validation
     }
+
+    private boolean hasReferencesExpanded(final PaymentInfo paymentInfo) {
+        return paymentInfo.getPayments().stream()
+                .allMatch(reference -> reference.getObj() != null);
```
katmatt: Here it would be nice if our `Reference` class would provide a `isExpanded` method, this would then allow us to use a method handle here: `.allMatch(reference::isExpanded)`. 

WDYT?

shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
```diff
@@ -16,21 +17,23 @@
...
 import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
-import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
 
+/**
+ * Deletes previous payments associated with the cart and adds the new one.
+ * {@code paymentInfo.payments} must be expanded in order to properly work, otherwise previous payments will not be removed.
+ */
```
katmatt: Great that you wrote javadoc. I especially like how you describe the behavior 👍

shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
```diff
@@ -16,21 +17,23 @@
...
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
-import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
 
+/**
+ * Deletes previous payments associated with the cart and adds the new one.
+ * {@code paymentInfo.payments} must be expanded in order to properly work, otherwise previous payments will not be removed.
```
katmatt: How about replacing '{@code...}' with `{@link PaymentInfo#getPayments()`? I prefer to use links because they add a bit more context. WDYT?

lauraluiz: Ah actually it was not a class, but the expansion path as it is expected by the platform :)
I will rewrite it to make it clear that I'm talking about paths.

shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
lauraluiz: Yes! I was actually thinking all the time "why don't we have a nice and useful method like that in the SDK??" 😄

shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
lauraluiz: In the end I joined both worlds :D

shopping-cart/app/com/commercetools/sunrise/framework/checkout/payment/DefaultCheckoutPaymentControllerAction.java
lauraluiz: https://github.com/commercetools/commercetools-jvm-sdk/issues/1422

PR #617, opened 2017-06-09
Disable matching variants

lauraluiz:
Disable it on all search calls and provide a way to enable it in POP (which is enabled by default to provide better results, but can be easily removed as suggested).


app/Module.java
```diff
@@ -106,6 +107,10 @@ protected void configure() {
         // Binding to truncate mini cart to fit it into limited session space
         bind(MiniCartViewModelFactory.class).to(TruncatedMiniCartViewModelFactory.class);
 
+        // Binding to enable matching variants on listing products
+        // IMPORTANT: comment the following line if your project does not require this functionality, leaving it on can severely affect performance
+        bind(ProductListFinder.class).to(ProductListFinderByCategoryWithMatchingVariants.class);
```
katmatt: That's a really nice example for the benefits of dependency injection 😄

PR #622, opened 2017-06-16
Provide configuration as default for wishlist, closes #621

### LINKED ISSUE #621: "wishlist.recentlyAddedItems" config value should come with a default
In general we need to try to avoid forcing developers to put mandatory configuration (there are exceptions, of course). But in this case it's clearly not necessary, we can set a default in `reference.conf`.

lauraluiz:


lauraluiz:
So I've removed it from config too. We have to think in the future how we put it in application.conf. Play provides all the reference.conf more or less copied. I was thinking too on just leaving it clear that they'll find all in reference.conf, to avoid confusion.
Let's see, for now I'll avoid duplication and just leave it there.

lauraluiz:
Closes #621

PR #626, opened 2017-06-20
Move wishlist controllers to separate packages.

katmatt:


wishlist/app/com/commercetools/sunrise/wishlist/content/viewmodels/WishlistPageContentFactory.java
```diff
@@ -49,11 +49,8 @@ protected void fillTitle(final WishlistPageContent viewModel, final ShoppingList
     }
 
     protected void fillItemsInTotal(final WishlistPageContent viewModel, final ShoppingList wishlist) {
-        if (wishlist != null) {
-            final List<LineItem> lineItems = wishlist.getLineItems();
-
-            viewModel.setItemsInTotal(lineItems == null ? 0 : lineItems.size());
-        }
+        final List<LineItem> lineItems = wishlist != null ? wishlist.getLineItems() : null;
+        viewModel.setItemsInTotal(lineItems == null ? 0 : lineItems.size());
```
katmatt: @lauraluiz I just found a small issue when visiting the wishlist for the first time: In this case `wishlist == null` and before my change, our template showed `content.itemsInTotal` instead of `0`. I fixed this bug wit the lines above.

lauraluiz: Good catch! 👍

PR #629, opened 2017-06-23
Improvements on final controllers

lauraluiz:


lauraluiz:
I'm confused where this PR comes from, but I'll just kill it.

PR #630, opened 2017-06-26
Improvements on final controllers

lauraluiz:


lauraluiz:
That changes from Starter :)

PR #635, opened 2017-07-12
Enable default CMS

lauraluiz:
Latest theme version expects logo from CMS

### DIFF
--- a/conf/cms/de/home.yaml
+++ b/conf/cms/de/home.yaml
@@ -1,63 +1,49 @@
-bannerTopLeft:
+banners[0]:
   title: Sommer Essentials
   subtitle: Die wichtigsten Stücke des Sommer Trends 2016
-  linkText: Jetzt bestellen
-  linkHref: men
-  imageSmall: /assets/img/cms/leather1_small-4to3.jpg
-  imageMedium: /assets/img/cms/leather1_medium-4to3.jpg
-  imageLarge: /assets/img/cms/leather1_large-4to3.jpg
+  linkLabel: Jetzt bestellen
+  link: /de/men
+  backgroundImage: /assets/img/cms/leather1-4to3.jpg
 
-bannerTopRight:
+banners[1]:
   title: Sommer Fashion
   subtitle: Beste Outfits für die Saison
-  linkText: Jetzt bestellen
-  linkHref: women
-  imageSmall: /assets/img/cms/woman1_small-4to3.jpg
-  imageMedium: /assets/img/cms/woman1_medium-4to3.jpg
-  imageLarge: /assets/img/cms/woman1_large-4to3.jpg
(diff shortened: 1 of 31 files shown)


lauraluiz:
This is how sunrise starter will look like by default. If they have a particular cms page for any of those controllers they just have to replace the "default". IMO we can also hide a "default" cmskey in some parent controller, but I'd prefer to be explicit about that (same as template name).

conf/cms/de/home.yaml
```diff
@@ -1,68 +1,58 @@
...
   title: Sommer Essentials
   subtitle: Die wichtigsten Stücke des Sommer Trends 2016
-  linkText: Jetzt bestellen
-  linkHref: men
-  imageSmall: /assets/img/cms/leather1_small-4to3.jpg
-  imageMedium: /assets/img/cms/leather1_medium-4to3.jpg
-  imageLarge: /assets/img/cms/leather1_large-4to3.jpg
+  linkLabel: Jetzt bestellen
+  link: /de/men
+  backgroundImage: /assets/img/cms/leather1-4to3.jpg
```
katmatt: Why did we have to update our file based cms config files? I don't see the connection to the other changes, may be I'm missing something here?

katmatt:
I think that being explicit about the CMS key is a good idea 👍

conf/cms/de/home.yaml
lauraluiz: Ah because before we used 3 different entries for images: small, medium, large.
But that disabled the possibility to use media capabilities of CMS, such as modifying size via query parameters (while keeping only one image entry on your content).
So we'll introduce passing options via handlebars helper and then each CMS implementation will define how these parameters are handled. That avoids entry duplication (imageSmall, imageMedium, imageLarge).
Of course once we have this new approach, we'll have to recover the images in different sizes, but for now I chose to removed them completely from the Theme to avoid confusion.

conf/cms/de/home.yaml
```diff
@@ -1,68 +1,58 @@
...
-  linkHref: women
-  imageSmall: /assets/img/cms/woman1_small-4to3.jpg
-  imageMedium: /assets/img/cms/woman1_medium-4to3.jpg
-  imageLarge: /assets/img/cms/woman1_large-4to3.jpg
+  linkLabel: Jetzt bestellen
+  link: /de/women
+  backgroundImage: /assets/img/cms/woman1-4to3.jpg
 
-bannerMiddle:
-  title: <img class="hidden-xs" src="/assets/img/cms/sun-logo.png"/><br/><br/>Here comes the sun
```
lauraluiz: Missing headerImage with this image.

conf/cms/de/home.yaml
katmatt: Now I remember that we already talked about it =)

PR #646, opened 2017-08-18
Fixing links issue

### LINKED ISSUE #639: Remove search parameters on changing category
Currently on changing to a category via faceted search, the query parameters such as pagination stays on the URL.

acbeni:
fixes #639

### DIFF
--- a/common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
+++ b/common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
@@ -43,8 +43,7 @@ public final class QueryStringUtils {
      * @return query string from the request
      */
     public static Map<String, List<String>> extractQueryString(final Http.Request httpRequest) {
-        return httpRequest.queryString().entrySet().stream()
-                .collect(toMap(Map.Entry::getKey, entry -> asList(entry.getValue())));
+        return extractQueryString(httpRequest, Collections.emptySet());
     }
 
     /**
@@ -55,9 +54,8 @@ public final class QueryStringUtils {
      */
     public static Map<String, List<String>> extractQueryString(final Http.Request httpRequest,final Set<String> ignoredParams) {
 
-        Set<String> ignored = Optional.ofNullable(ignoredParams).orElseGet(HashSet::new);
         return httpRequest.queryString().entrySet().stream()
-                .filter(stringEntry -> !ignored.contains(stringEntry.getKey()))
+                .filter(stringEntry -> !ignoredParams.contains(stringEntry.getKey()))
                 .collect(toMap(Map.Entry::getKey, entry -> asList(entry.getValue())));
     }
 
--- a/common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
+++ b/common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
@@ -5,9 +5,7 @@ import play.mvc.Http;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 
 import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.joining;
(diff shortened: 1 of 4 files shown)


product-catalog/app/com/commercetools/sunrise/productcatalog/productoverview/search/facetedsearch/categorytree/viewmodels/CategoryTreeFacetOptionViewModelFactory.java
```diff
@@ -25,15 +25,24 @@
     private final List<Locale> locales;
     private final CategoryTree categoryTree;
     private final ProductReverseRouter productReverseRouter;
+    private final Set<String> ignoredParams;
```
lauraluiz: Why not make it static better and declare it directly in here?
```java
private static final Set<String> IGNORED_PARAMS = Collections.singleton("page"); 
```
It saves some lines of code and it's a bit more efficient.

common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
```diff
@@ -50,6 +48,20 @@ private QueryStringUtils() {
...
 
     /**
+     * Extracts the query string from the request.
+     * @param httpRequest current HTTP request
+     * @param ignoredParams params that should not be included in the link
+     * @return query string from the request
+     */
+    public static Map<String, List<String>> extractQueryString(final Http.Request httpRequest,final Set<String> ignoredParams) {
+
+        Set<String> ignored = Optional.ofNullable(ignoredParams).orElseGet(HashSet::new);
```
lauraluiz: In Sunrise we follow the rule that unless a parameter is annotated with `@Nullable`, it can't be null. Therefore this line of code is unnecessary.

common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
```diff
@@ -50,6 +48,20 @@ private QueryStringUtils() {
...
+        return httpRequest.queryString().entrySet().stream()
+                .filter(stringEntry -> !ignored.contains(stringEntry.getKey()))
+                .collect(toMap(Map.Entry::getKey, entry -> asList(entry.getValue())));
```
lauraluiz: To avoid duplication, I would let the other `extractQueryString` method to call this one with an empty set as `ignoredParams` (i.e. `Collections.emptySet()`). But I'm not sure how efficient would be the stream with a filter predicate that it's always false. Otherwise what we can surely avoid duplicating is the Collector `toMap(Map.Entry::getKey, entry -> asList(entry.getValue()))`. That would be enough to avoid duplication, I leave it to your decision :)

common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
```diff
@@ -50,6 +48,20 @@ private QueryStringUtils() {
     }
 
     /**
+     * Extracts the query string from the request.
+     * @param httpRequest current HTTP request
+     * @param ignoredParams params that should not be included in the link
+     * @return query string from the request
+     */
+    public static Map<String, List<String>> extractQueryString(final Http.Request httpRequest,final Set<String> ignoredParams) {
```
lauraluiz: We don't have many tests in Sunrise yet, but we actually have one for this class 😅 
Could you write some tests for this method?

product-catalog/app/com/commercetools/sunrise/productcatalog/productoverview/search/facetedsearch/categorytree/viewmodels/CategoryTreeFacetOptionViewModelFactory.java
acbeni: 👍 good idea.
i will include t in the code review.

common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
acbeni: ok if thats the case it will keep the code cleaner 👍

common/app/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtils.java
acbeni: 😅 didnt know we had a test for that one. 😄 now the test is added.

acbeni:
@lauraluiz the code review has been done 👍

common/test/com/commercetools/sunrise/framework/viewmodels/forms/QueryStringUtilsTest.java
```diff
@@ -101,4 +102,17 @@ public void extractsQueryString() throws Exception {
...
                         .isEqualTo(queryString));
     }
+
+    @Test
+    public void extractsQueryStringIgnoringAParameter() throws Exception {
+
+        final String IGNORED_PARAM = "foo";
+        final Map<String, List<String>> queryString = someQueryString();
+
+        queryString.remove(IGNORED_PARAM);
```
lauraluiz: If you remove the parameter it won't be there for the test and it will always pass ;)

acbeni: 👍 good catch
corrected.

PR #657, opened 2017-09-06
Fix encoding of selectable attributes.

### LINKED ISSUE #649: Selectable attributes fails for numbers
Having a product attribute combination of string + number seems to fail for -at least- a secondary attribute type. For example: "1stClass" + "12.0".

The generated cross attributes is "1stClass-12" and the value in the select option is "1stClass" and "12". But the printed number is "12.0". Maybe it tries to match the label instead of the value?

katmatt:
Closes #649


common/test/com/commercetools/sunrise/framework/viewmodels/content/products/SelectableProductAttributeViewModelFactoryTest.java
```diff
@@ -0,0 +1,122 @@
...
+        when(fakeProductType.findAttribute(NUMBER_ATTRIBUTE_NAME)).thenReturn(Optional.of(numberAttributeDefinition));
+
+        numberAttribute1 = Attribute.of(NUMBER_ATTRIBUTE_NAME, AttributeAccess.ofDouble(), 12.0);
+        numberAttribute2 = Attribute.of(NUMBER_ATTRIBUTE_NAME, AttributeAccess.ofDouble(), 13.0);
+
+        when(fakeProductType.getId()).thenReturn(PRODUCT_TYPE_ID);
+        when(fakeProductType.toReference()).thenReturn(productTypeRef);
+        when(fakeProductType.getAttributes()).thenReturn(Arrays.asList(stringAttributeDefinition, numberAttributeDefinition));
+        productTypeRef = Reference.ofResourceTypeIdAndObj(ProductType.resourceTypeId(), fakeProductType);
+        when(fakeProductTypeLocalRepository.findById(PRODUCT_TYPE_ID)).thenReturn(Optional.of(fakeProductType));
```
lauraluiz: This looks like a lot of work for just one test. My biggest fear is that it might not be easily scalable for more tests, and we are in fact missing many tests cases.  In the end I think we should have one test where it is clear what is encoded and what is not (with a single attribute) and then try with combination of attributes and edge cases (many attributes-single value, single attribute-multiple value).
Also another suggestion would be to use JSON representations in separate files of a real responses from the platform, and having them in each test.

common/test/com/commercetools/sunrise/framework/viewmodels/content/products/SelectableProductAttributeViewModelFactoryTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.*;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for {@link SelectableProductAttributeViewModelFactory}.
+ */
+public class SelectableProductAttributeViewModelFactoryTest extends WithApplication {
```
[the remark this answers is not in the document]
katmatt: I use the app to get a SelectableProductAttributeViewModelFactory instance. And yes, it should be located under pt, but it seems to work in its current location too. Which is suspicious... 😉

common/test/com/commercetools/sunrise/framework/viewmodels/content/products/SelectableProductAttributeViewModelFactoryTest.java
katmatt: I agree with you, let's see if we can find a better solution together!

katmatt:
@lauraluiz I improved the test and I"m looking forward to your feedback 😄

lauraluiz:
Let's merge them already. Further improvements can be done once we split that class into smaller classes. Good job! 👍

PR #659, opened 2017-09-08
Forget password (pre-1.0.0)

lauraluiz:


activator
```diff
@@ -1,334 +0,0 @@
-#!/usr/bin/env bash
```
katmatt: Nice that you removed the activator related files 👍

app/demo/myaccount/RecoverPasswordController.java
```diff
@@ -0,0 +1,8 @@
+package demo.myaccount;
```
katmatt: Would be nice to have the `demo` package for the app on master too. What was the reason to drop it on master?

app/email/smtp/EmailSenderProvider.java
```diff
@@ -0,0 +1,40 @@
+package email.smtp;
```
katmatt: Don't think it's worth the effort, but it would be more consistent to move the `email` package into the `demo` package ;-)

activator
lauraluiz: Yep, we didn't need them anymore since nobody is supposed to run this app.

app/demo/myaccount/RecoverPasswordController.java
lauraluiz: The package name was simply renamed, as the convention for Play applications is to have the controllers inside the `controllers` folder. In our case "demo" could made some sense, but once imported this to Starter we had to rename it to the convention because it was meant for a normal Play project.

katmatt: Ah, I forgot about this convention and thanks for your explanation 😄

app/email/smtp/EmailSenderProvider.java
lauraluiz: I see your point, but if `demo` is actually `controllers` then that's the reason why I kept the email settings/providers in their own package as in M6 (not controllers). So I'm just assuming `demo` is actually called `controllers` all the time.

PR #678, opened 2017-10-12
SUPPORT-1511: Add the field ID To the errorMessage.

ahalberkamp:
For CDL [1] I have to rework the error handling for forms, so that the error message, which says that a required value is not set, is displayed directly "under" the empty form field. For doing this I need a kind of a mapping between the error message and the form field, to which the error message belongs. 
This mapping cannot be done using the sunrise Framework right now, because its Errorbeans, which encapsulate the errors, only contains the localized error message but not the ID of field, to which the error belongs (please see: [2])


common/test/com/commercetools/sunrise/common/template/engine/handlebars/PlayJavaFormResolverTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PlayJavaFormResolverTest {
+
+    PlayJavaFormResolver formResolver = new PlayJavaFormResolver(singletonList(Locale.ENGLISH), new ErrorFormatter() {
+        @Override
+        public String format(List<Locale> locales, String message) {
+            return message;
+        }
+    });
```
lauraluiz: Suggestion: the instantiation of ErrorFormatter can be a bit smaller with:
```java
new PlayJavaFormResolver(singletonList(Locale.ENGLISH), (locales, message) -> message);
```

common/test/com/commercetools/sunrise/common/template/engine/handlebars/PlayJavaFormResolverTest.java
```diff
@@ -0,0 +1,52 @@
...
+    public void extractErrors() throws Exception {
+        String errorField1 = "field1";
+        String errorField2 = "field2";
+        Form form = Mockito.mock(Form.class);
+        Map<String, List<ValidationError>> errorMap = new HashMap<String, List<ValidationError>>();
+        errorMap.put(errorField1, Arrays.asList(new ValidationError("errorkey1", "errorMessage1"),
+                new ValidationError("errorkey2", "errorMessage2")));
+        errorMap.put(errorField2, Collections.singletonList(new ValidationError("errorkey21", "errorMessage21")));
+
+        Mockito.when(form.errors()).thenReturn(errorMap);
```
lauraluiz: I would extract all this:
```java
Form form = Mockito.mock(Form.class);
Map<String, List<ValidationError>> errorMap = new HashMap<String, List<ValidationError>>();
errorMap.put(errorField1, Arrays.asList(new ValidationError("errorkey1", "errorMessage1"),
                new ValidationError("errorkey2", "errorMessage2")));
errorMap.put(errorField2, Collections.singletonList(new ValidationError("errorkey21", "errorMessage21")));
Mockito.when(form.errors()).thenReturn(errorMap);
```
into a method, called for example `formWithSomeErrorsForFields(field1, field2)`
This way the test is keeping only the important parts, with all boilerplate elsewhere.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.22.0"
+version in ThisBuild := "0.23.0"
```
lauraluiz: The version should not be modified manually, it is automatically increased on the release process, so we need to revert this change :)

lauraluiz: Actually never mind this one, I can solve it via the conflict resolver of Github :)

common/test/com/commercetools/sunrise/common/template/engine/handlebars/PlayJavaFormResolverTest.java
ahalberkamp: 694feaf

common/test/com/commercetools/sunrise/common/template/engine/handlebars/PlayJavaFormResolverTest.java
ahalberkamp: 694feaf

PR #679, opened 2017-10-12
Fix range facets

lauraluiz:
@katmatt this is the fix for the range facets. I've tested them manually carefully, but we still need the automated tests. As they are still not exposed, I would like to wait for the backend to fix the issues before really testing and documenting them, but at least I wanted to fix this bug. The related ticket is here: https://github.com/commercetools/commercetools-sunrise-java/issues/452

### DIFF
--- a/common/app/com/commercetools/sunrise/search/facetedsearch/RangeUtils.java
+++ b/common/app/com/commercetools/sunrise/search/facetedsearch/RangeUtils.java
@@ -8,6 +8,7 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 
 import javax.annotation.Nullable;
+import java.math.BigDecimal;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


common/app/com/commercetools/sunrise/search/facetedsearch/RangeUtils.java
```diff
@@ -89,17 +98,48 @@ private RangeUtils() {
...
+    private static <T extends Comparable<T>> boolean hasNoBounds(final FilterRange<T> range) {
+        return range.lowerBound() != null || range.upperBound() != null;
+    }
+
+    /**
+     * Cleans the decimal part of the endpoint value due to the backend answering with decimals but not accepting them.
+     * For example, the backend might answer with 9002.0, but it needs to be converted to 9002.
+     * @param endpointValue the value of the endpoint to be cleaned
+     * @return the value of the endpoint without the decimal part, in case the value was a number
+     */
```
katmatt: Great explanation 👍

PR #684, opened 2017-10-24
Accept i18n errors

lauraluiz:
All error messages can be localized now.
The ones defined in Sunrise are located in the bundle "messages" (as shows the `messages.yaml` files). The ones pre-defined by Play (i.e. https://github.com/playframework/playframework/blob/2.5.18/framework/src/play/src/main/resources/messages.default) are located in the default bundle ("main" if not changed). They can also be localized using the default Play way of creating messages files for each language.


common/app/com/commercetools/sunrise/common/utils/ErrorFormatter.java
```diff
@@ -18,16 +21,22 @@
      * @param message error message
      * @return the error message localized and formatted
      */
-    String format(final List<Locale> locales, final String message);
+    String format(final List<Locale> locales, final String message, final Map<String, Object> hashArgs);
```
katmatt: I would prefer the parameter name `args` here.

common/test/com/commercetools/sunrise/common/template/engine/handlebars/PlayJavaFormResolverTest.java
```diff
@@ -35,7 +32,7 @@ public void extractErrors() throws Exception {
 
     private void checkError(ErrorBean error, String field, String key, String message) {
```
katmatt: I didn't check it in the IDE, but to me it looks the `key` parameter is unused?

common/test/com/commercetools/sunrise/common/utils/ErrorFormatterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Collections.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ErrorFormatterImplTest {
```
katmatt: I like the new tests 👍

common/app/com/commercetools/sunrise/common/utils/ErrorFormatter.java
lauraluiz: I named them `hashArgs` here and in `I18nResolver` because they are hash arguments. What don't you like in particular from the name?

lauraluiz: Done!

acbeni:
@lauraluiz i like what has been done in general, there is major quality amelioration that can be done and that i think is worth the effort, which is to work with generated classes like the android `R.java` which would allow us to get hold of all the messages in type safe way.

WDYT?

lauraluiz:
Having type-safe messages sounds really interesting @acbeni ! I hadn't thought of that solution. Could you please create an issue for that? I would address that after v1.0.0 though.

lauraluiz:
And actually all theme resources might benefit from it.

acbeni:
@lauraluiz i created issue to follow up #685

PR #695, opened 2017-12-07
Fix pagination total pages

lauraluiz:
Fixed an error reported in which the last page was usually not listed due to division of `Long` returning a non-decimal value, therefore `ceil` method was not able to round up.
Also provided tests and cleaned up a bit the other tests.
Will release in a fix version as soon as it's merged.

### DIFF
--- a/common/app/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactory.java
+++ b/common/app/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactory.java
@@ -121,11 +121,11 @@ public abstract class AbstractPaginationViewModelFactory extends ViewModelFactor
         return Math.min(currentPage + paginationSettings.getDisplayedPages(), totalPages);
     }
 
-    private long calculateTotalPages(final PagedResult<?> pagedResult, final long currentPage) {
+    static long calculateTotalPages(final PagedResult<?> pagedResult, final long currentPage) {
         if (pagedResult.isLast()) {
             return currentPage;
         } else if (pagedResult.getCount() > 0) {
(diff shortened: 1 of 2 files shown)


common/app/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactory.java
```diff
@@ -121,11 +121,11 @@ private long calculateTopThresholdPage(final long currentPage, final long totalP
         return Math.min(currentPage + paginationSettings.getDisplayedPages(), totalPages);
     }
 
-    private long calculateTotalPages(final PagedResult<?> pagedResult, final long currentPage) {
+    static long calculateTotalPages(final PagedResult<?> pagedResult, final long currentPage) {
```
katmatt: You had to drop the `private` modifier for testing?

common/pt/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactoryTest.java
```diff
@@ -142,55 +165,60 @@ public void calculatesPaginationForSeventhPage() throws Exception {
                 .containsExactly(false, false, true, false, false, false);
     }
 
+    @Test
+    public void calculatesTotalPagesForMiddlePage() throws Exception {
```
katmatt: I like how these tests read 👍

common/app/com/commercetools/sunrise/search/pagination/viewmodels/AbstractPaginationViewModelFactory.java
lauraluiz: Yep

